PR #43, opened 2023-02-10
Enable Clippy on pushes/PRs to main

philpax:
In preparation for PRs from other folks - run Clippy on PRs. This also 'fixes' the remaining deprecations; I've added more reasoning to my changes in each commit.


philpax:
TODO:

- [ ] Move Clippy to the front so it fails fast and hard
- [ ] Set up https://github.com/swellaby/rusty-hook so that it's easy for people to set up pre-push hooks
- [ ] See if it's possible to allow the use of `todo` iff it has a descriptive argument
- [ ] Allow deprecations

FredrikNoren:
@philpax So how does this work now exactly? Is the idea that it runs clippy automatically locally on `git push`? What do people need to do to make that work?

philpax:
Nothing, it should just work. rusty-hooks gets pulled in as a dependency and automatically sets up. Give it a shot and lmk if it works for you

PR #44, opened 2023-02-11
Try a different cargo cacher

FredrikNoren:


FredrikNoren:
Still far from great.. but bevy is switching to this cacher and doesn't seem to be worse so might as well merge

PR #45, opened 2023-02-11
Build on stable

### LINKED ISSUE #31: Build on stable Rust
At present, the runtime does not build on stable Rust. This is due to the use of four features:

```
#![feature(map_many_mut)]
#![feature(cursor_remaining)]
#![feature(coerce_unsized)]
#![feature(unsize)]
```

The first two features are primarily used for convenience and can be replaced with slightly-less-ergonomic equivalents. The latter two features are used to enable easy construction of `Cb` callback wrappers - it's likely that the downstream code will need to be changed to not rely on this, or another way is found to make `Cb` construction simpler.

philpax:
Fixes #31.

Had to remove the four features in use. `cursor_remaining` was simple to do; `map_many_mut` required some rearranging of one function - but the biggest was removing the use of `unsize` and `coerce_unsized` for `Cb`.

To do this, I tried out a few different solutions, but settled on changing `Cb` to be an alias for `Arc<Debuggable<T>>`, and created `cb` and `cb_arc` to create `Cb`s (analogous to `Cb::new` and `Cb(/* some arc */)`) respectively. It would have been nice to do it with a newtype as before, but type inference appears to fail in that case.

This now builds on stable Rust 1.67.0, no tricks involved. Probably builds on versions earlier than that, too, but not by much.

Mild annoyance: stable `rustfmt` does not support the import formatting settings, so the imports will be reformatted to the standard (as you can see in the files that were edited)

### DIFF
--- a/app/src/main.rs
+++ b/app/src/main.rs
@@ -8,10 +8,12 @@ use kiwi_debugger::RendererDebugger;
 use kiwi_ecs::{EntityData, SystemGroup, World};
 use kiwi_element::{element_component, Element, ElementComponentExt, Hooks};
 use kiwi_network::{
-    client::{GameClient, GameClientNetworkStats, GameClientRenderTarget, GameClientServerStats, GameClientView, UseOnce}, events::ServerEventRegistry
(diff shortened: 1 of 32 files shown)


FredrikNoren (approved):
Awesome!

PR #49, opened 2023-02-12
fix: ServeDir should use `nest_service` instead of `route`

### LINKED ISSUE #47: Examples Fail To Load Assets
Running `kiwi run` from within any example's directory that uses assets leads to errors in the log despite the corresponding asset's `main.json` existing in `target/assets` (note that many permutations of the path were tried, and failed):
```
[2023-02-12T17:35:04Z WARN  kiwi_std::uncategorized::download_asset] Request for http://192.168.1.66:8999/assets/castle.glb/objects/main.json failed: 404
[2023-02-12T17:35:04Z ERROR kiwi_object] Failed to load object: Downloading http://192.168.1.66:8999/assets/castle.glb/objects/main.json failed, bad status code: 404
```

SK83RJOSH:
This fixes #47, Mithun helped me track down the cause 👍

### DIFF
--- a/app/src/server/mod.rs
+++ b/app/src/server/mod.rs
@@ -86,7 +86,7 @@ pub const QUIC_INTERFACE_PORT: u16 = 9000;
 fn start_http_interface(runtime: &tokio::runtime::Runtime, project_path: &Path) {
     let router = Router::new()
         .route("/ping", get(|| async move { "ok" }))
-        .route("/assets", get_service(ServeDir::new(project_path.join("target"))).handle_error(handle_error))
+        .nest_service("/assets", get_service(ServeDir::new(project_path.join("target"))).handle_error(handle_error))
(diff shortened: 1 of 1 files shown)


philpax:
This one's on me - I made it typecheck, but I didn't realise that it wouldn't route properly. I'll see if the Axum folks have an issue about this footgun (imo)

FredrikNoren:
Wait, this will make it serve all assets under root right? I don't think that's what we want; this http interface will probably have more endpoints in the future. We should probably just change the name of the endpoint to something other than `assets` so it's clearer.

philpax:
I believe it was serving all assets under root before; I updated Axum last week and made the smallest change to make it compile and confirmed the examples still worked for me (which I'm suspecting is because I already had them on disk), but that change resulted in the routing being wrong.

`nest_service` is from the Axum example https://github.com/tokio-rs/axum/blob/main/examples/static-file-server/src/main.rs so it should be the correct behaviour.

FredrikNoren:
@philpax Ah ok nice

PR #79, opened 2023-02-13
ci: Restore timestamps from git

FredrikNoren:


FredrikNoren:
Seems this method wont work; build broke on non-code change

FredrikNoren:
This seems to work and seems ~5min faster than the baseline. I still don't quite trust it though so if we get weird build errors this might be the culprit.

PR #84, opened 2023-02-14
fix: Collider::reverse_indices default now false

SK83RJOSH:
I don't believe indices need reversed at all. In my testing this has detrimental effects to every object tested; so I'm changing the default to false.

@FredrikNoren I think flip_normals ends up having the same effect as this, and in addition to that, both are probably not needed from my testing so we may want to drop them (it seems our winding order matches PhysX; your original issue with more details on this issue is long gone so I'm basing this on my observations with ~12 different objects).

### DIFF
--- a/crates/build/src/pipelines/models/mod.rs
+++ b/crates/build/src/pipelines/models/mod.rs
@@ -137,7 +137,7 @@ pub enum Collider {
     FromModel {
         #[serde(default)]
         flip_normals: bool,
-        #[serde(default = "true_value")]
+        #[serde(default)]
         reverse_indices: bool,
     },
     Character {


SK83RJOSH:
Actually scratch that! This does actually have a positive effect on collisions, but causes things to be inverted in the PVD. I'm going to leave this open for now, but we should investigate this at some point - because this makes PVD quite unusable.

FredrikNoren:
Uh yeah weird. We added the reverse_indices to get some other colliders to work. I agree that it feels like flip_normals should do the same, but I also suspect that we tested that first when we added this, so we probably need to verify that the other colliders work with just that flag before we remove reverse_indices (like big medium small, quixel etc.)

SK83RJOSH:
Agreed! I'm probably not going to go too far down this rabbit hole until the weekend; but I definitely want to figure out why we need to change the indices at all. Based on the PVD rendering getting messed up I think there's a bug somewhere on our side that's leading to weird behavior with mesh/triangle shapes. When/if I figure it out I'll update this and run some more thorough tests.

philpax:
@SK83RJOSH any update?

ten3roberts:
We are closing this as of now as the issue has gone stale and we are not sure if this is an actual issue.

If we experience problems with colliders in the future we will reopen this

PR #99, opened 2023-02-15
feat: add contact/rest offset components

SK83RJOSH:


SK83RJOSH:
Starting to see that it's going to be pretty tedious to keep adding these in two places. @FredrikNoren any ideas for how we might be able to rewrite the systems to apply these changes to the bodies both on the component being dirty and the initial creation of the body? :slightly_smiling_face:

FredrikNoren (approved):

PR #103, opened 2023-02-16
ci: Test not in release

FredrikNoren:


FredrikNoren:
This seems to bring down build times for a dummy change to ~23min.

PR #140, opened 2023-02-21
feat: Add minigolf example

SK83RJOSH:


philpax (approved):
lgtm, you're already working on removing `State` so won't mention that. `pipeline.json` can now also be used without the surrounding array if you want to make that a little cleaner

FredrikNoren (approved):
Awesome! lgtm, only comment I had was also getting rid of `State`, but sounds like you guys are already on top of that.

SK83RJOSH:
Okay well... I removed the player state and everything works... for about 30 seconds before the server hangs... 😢

Issue #150, opened 2023-02-22
Explain the network protocol and its current limitations


philpax:
The guide should have a document explaining the current state of the network protocol, what our next steps are, and what the current limitations are.

Here's a quick list of facts that can be used to write up a more complete document:

- We currently only support desktop clients, and use QUIC through the `quinn` library as our communication protocol.
- We are actively working on web deployments. To begin with, we may use WebSockets (due to their support across all browsers), but we would like to adopt WebTransport as soon as possible. This is dependent on browser support and `wasm-bindgen` availability.
---
- The HTTP (TCP) port is `8999` and the QUIC (UDP) port is `9000`.
---
- We synchronise all entities with at least one component marked with the `Networked` attribute. Only the components marked `Networked` will be sent to the client.
- Most core components are `Networked`. This may have unintended ramifications in terms of cheating, especially for hostile clients.
- By default, custom components are not `Networked`. This is something users have to opt into.
- The client is fundamentally designed around runtime flexibility of logic, which is non-ideal for avoiding cheaters. Further R&D is required, but it is likely that there is no silver bullet and the solution will be game-dependent.
---
- All gameplay logic is currently server-authoritative.
- We currently do not have any form of latency-hiding, including prediction, rollback, or clientside logic.
- We previously had rollback, but removed it due to its relative inflexibility.
- Our plan is to introduce clientside and shared logic that can be used for user-defined prediction with runtime assistance, but this will take some time.

xunilrj:
I am interested in helping with this. @philpax

ggambetta:
Happy to brainstorm about client-side prediction and server reconciliation :) I've [written a bit](https://gabrielgambetta.com/client-server-game-architecture.html) about this.

philpax:
> I am interested in helping with this. @philpax

Go for it! Fork the repo, add a Markdown file to docs/src, and add a link to it under the `SUMMARY.md` Reference section. It's just mdbook, so you should be able to work on it locally.

Once you get started, I'll assign the issue to you 👌

canadaduane:
Have you considered using WebRTC at first, for its ability to support UDP? Here is an example implementation of a javascript server & client: https://geckos.io/

arthuro555:
For UDP on the web, geckos has a few problems that make it neat for small projects but unpractical at scale, notably requiring all the ports of a machine to be open, not working in docker/kubernetes...
See https://github.com/geckosio/geckos.io/issues/245

[...]

PR #157, opened 2023-02-23
(WIP) chore: doc about network

xunilrj:
This PR creates documentation about the network protocol.

The style is still confusing in the sense that is a mix of a code walk-through, reference documentation, and a tutorial. We can decide later which style to keep, or to split into multiple pages.

Closes https://github.com/AmbientRun/Ambient/issues/150

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -14,6 +14,7 @@
 - [Runtime](./runtime.md)
 - [Project](./project.md)
 - [Asset pipeline](./asset_pipeline.md)
+- [Networking](./networking.md)
 - [Terminology](./terminology.md)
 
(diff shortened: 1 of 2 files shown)


philpax:
Hey, super sorry about the time it's taken to look at this. It's been a busy two weeks.

Thanks for the contribution! We're trying to keep the documentation user-facing (that is, "developers building projects against Ambient"), so our preference is to not mention too much about the internals or how it's implemented.

Would you be able to revise this to discuss the user-facing experience (i.e. using the `Networked` attribute in `ambient.toml`), and to discuss the high-level points from the issue? The most important stuff is about the network architecture and its current limitations - e.g. no rollback, etc.

Issue #153, opened 2023-02-23
ci: Linux builds growing for each new build


FredrikNoren:
Something strange is going on with the ci; it seems like the Linux build cache gets bigger and bigger with each new build. I haven't been able to figure out why yet. It seems like Windows and Mac works fine.

philpax:
Do we have any information about the contents of the cache?

FredrikNoren:
No I was trying to see if there's a way to download them but couldn't find any, think github doesn't allow it. Was playing around with saving it to artifacts here but didn't look at it yet: https://github.com/AmbientRun/Ambient/pull/125

philpax:
I'm about to bust the Linux cache to allow #156 to build, so we'll need to collect data again.

daniellavoie:
I've customized the build on my own fork to generate a `tree target` output.

https://github.com/daniellavoie/Ambient/blob/debug-cache/cache.log

I'm currently running another build and will git diff to see if consecutive builds creates new entries. Meanwhile, the output should still help figuring out if there is any suspicious file out there.

daniellavoie:
Just found out `tree --du -h` will rerun the same flow with this command to get a better idea of what's what.

daniellavoie:
git diff between two builds on the same commit. doesn't. Running again with `tree --du-h`.

```
git diff
diff --git a/cache.log b/cache.log
index 354e187..b049d80 100644
--- a/cache.log
+++ b/cache.log
@@ -761,7 +761,7 @@ target
 │   │   │   │   ├── 643449dc0c65ac0e-ScNPhaseCore.o
 │   │   │   │   ├── 643449dc0c65ac0e-ScPhysics.o
 │   │   │   │   ├── 643449dc0c65ac0e-ScRigidCore.o
-│   │   │   │   ├── 643449dc0c65ac0e-ScRigidSim.o
+│��  │   │   │   ├── 643449dc0c65ac0e-ScRigidSim.o
 │   │   │   │   ├── 643449dc0c65ac0e-ScScene.o
 │   │   │   │   ├── 643449dc0c65ac0e-ScShapeCore.o
 │   │   │   │   ├── 643449dc0c65ac0e-ScShapeInteraction.o
@@ -1077,7 +1077,7 @@ target
 │   │   │   ├── out
 │   │   │   ├── output
 │   │   │   ├── root-output
-│   │   │   ��── stderr
+│   │   │   └── stderr
[...]

PR #170, opened 2023-02-24
chore(ci): trace target directory content for cache troubleshooting

daniellavoie:
#153 reported a suspicious cache usage increase.

This PR introduces a new step in the build process to output the content of the `target` directory after the build. This will help investigate if something is wrong builds after builds.

I ran multiple builds with this change and couldn't observe any file change or file increase from the same commits.

Here is an example of the new step in action: https://github.com/daniellavoie/Ambient/actions/runs/4257943674/jobs/7408628309

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install build dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

FredrikNoren:
@daniellavoie Yup good idea, merging.

PR #178, opened 2023-02-26
feat: update the window title with the project name - fixes #171

### LINKED ISSUE #171: Update the client's window title with the server's project name
This is a minor quality of life feature, but might be a good challenge for someone who wants to get to know the codebase.

To do this, you'll need to retrieve the `project.name` from the manifest on the server: https://github.com/AmbientRun/Ambient/blob/676012b96f6f839879cbf2cd280e3994ed008ca5/app/src/server/mod.rs#L63

(I suggest storing it as a component on the resource entity.)

You'll then need to send it to the client during the handshake: https://github.com/AmbientRun/Ambient/blob/676012b96f6f839879cbf2cd280e3994ed008ca5/crates/network/src/protocol.rs#L86

You will need to do some plumbing to get the project name to there.

This will then be accessible in `ClientInstance::run`: https://github.com/AmbientRun/Ambient/blob/676012b96f6f839879cbf2cd280e3994ed008ca5/crates/network/src/client.rs#L373

This is called by `GameClientView` above. Replicate the `set_connection_status` state hook, and pass in `set_project_name` to `ClientInstance` and call that once the project name is available.

In `GameClientView`, update the `window_title` component with `project_name`, similar to how it's done here: https://github.com/AmbientRun/Ambient/blob/676012b96f6f839879cbf2cd280e3994ed008ca5/app/src/client/mod.rs#L36

The `GameClientView::render` method is called every time its hook state changes, so you don't need to do anything special.

Note that there are *two* worlds on the client - the primary "UI" world (which is what the majority of the `ElementComponent`s dealing with UI use), and the game world kept in `ClientGameState`. Make sure you're updating `window_title` on the former, not the latter.

MavethGH:
This PR adds a new `ServerData` struct that is transmitted from the server to the client during the handshake, and is intended to store various metadata that might be relevant to the client. For now all it stores is the project name, but that can be easily changed in the future.

### DIFF
--- a/app/src/server/mod.rs
+++ b/app/src/server/mod.rs
@@ -6,7 +6,7 @@ use std::{
     time::SystemTime,
 };
 
-use ambient_core::{app_start_time, asset_cache, dtime, no_sync, time};
+use ambient_core::{app_start_time, asset_cache, dtime, no_sync, project_name, time};
 use ambient_ecs::{world_events, ComponentDesc, ComponentRegistry, EntityData, Networked, SystemGroup, World, WorldStreamCompEvent};
 use ambient_network::{
     bi_stream_handlers, datagram_handlers,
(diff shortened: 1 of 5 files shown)


philpax:
Awesome, thank you! The only changes I made were to:
- Add `Debuggable` to `proejct_name`, so that its value could be seen in the ECS dump
- Remove the space before the newline in the description
- Run `cargo run -- update-interface-components` to regenerate `ambient.toml`

Aside from that, that was perfect. Thanks for the PR! 😁

Issue #39, opened 2023-02-10
Can we support `cargo run` for Ambient projects?


FredrikNoren:
We _might_ be able to support `cargo run` in a Ambient project folder; we need to add this to `Cargo.toml` (for example in the examples):
```
[target.wasm32-wasi]
runner = "ambient"
```
To make this work we would also need to make the asset build part of the cargo build. Needs to be experimented with a bit though.

philpax:
I've been thinking about this more and am starting to think that we should flip the compilation model around so that you can run *any* arbitrary WASM module, and we remove Rust compilation entirely from the asset pipeline. That is, `cargo run` builds your WASM, which is then passed as an input to `kiwi run`, and you can do this with any compiler.

The thing that I haven't figured out is how this works with dependencies. When those dependencies have code of their own, what happens? Do we make the user go in and build those manually? Do we enforce that dependencies must have WASM pre-built? What if you don't have the relevant toolchain?

PR #190, opened 2023-03-01
Wasmtime update and clientside WASM

### LINKED ISSUE #1: Update wasmtime / wit-bindgen version
At the time of creating this issue, we are using a relatively old (Sept 2022) version of wasmtime. We have been unable to update due to the ongoing work with wit-bindgen.

However, it looks like this work is soon coming to an end: https://github.com/bytecodealliance/wasm-tools/pull/867

We should hopefully be able to update soon.

philpax:
Fixes #1 and #6.

Still to do:

- [x] Wire up Wasmtime in `ambient_wasm`. I've tested this works with a separate sandbox that stubbed out all of the exports, but now I need to actually wire it up to Ambient.
- [x] Rename `component::ComponentType` to `component::Value`
- [x] Check the docs on the guest still make sense.
- [x] Figure out how to accommodate future client bindings. Checking in with the Bytecode Alliance [here](https://bytecodealliance.zulipchat.com/#narrow/stream/327223-wit-bindgen/topic/Best.20way.20to.20share.20generated.20types.20amongst.20worlds).
- [x] Deal with the excessive compilation times from building the same package with two different features. Maybe supply a different `--target-dir`?
- [x] Bytecode transmission to the client is slow. The unoptimised debug WASM bytecode files are 14MB, and Quinn doesn't much like that. Should find a way to expose this over the asset interface.
- [x] Consider a way to opt out of client builds to avoid "paying the cost" of clientside WASM when you don't need it?
- [x] Fix the physics example (fails with a component error)
- [x] Use published versions of crates where applicable


philpax:
So I have the `async` example working, but the `physics` example explodes quite dramatically: it looks like `entity.spawn` is not marshalling the list of tuples across correctly. I suspect this might be a mismatch between the wit-bindgen version and the wasmtime version; more investigation is required.

The other thing is that this now requires the component model, so it can't just consume wasm output by rustc directly. It needs to be converted to a component and have an adapter for wasi preview1 -> preview2 injected:
```
wasm-tools component new ./target/wasm32-wasi/debug/physics.wasm -o examples/physics/build/physics.wasm --adapt wasi_snapshot_preview1.wasm
```
To test, I've disabled the automatic build of Rust in the build script, but I need to think about the best way to handle this going forward. I'd like to take `rustc` out of the build pipeline entirely, so this might be a good time to do so... but that does break the nice `ambient run` workflow. Could be a chance to tie in #39, but there are three steps: build, componentize, run.

I could have the build process componentize your raw WASM for you, but that seems a little finicky + could be something the user doesn't want (especially if they're using somebody else's components).

philpax:
> So I have the `async` example working, but the `physics` example explodes quite dramatically: it looks like `entity.spawn` is not marshalling the list of tuples across correctly. I suspect this might be a mismatch between the wit-bindgen version and the wasmtime version; more investigation is required.

I've updated the wasmtime version and the issue still persists. I've opened an issue with `wit-bindgen` [here](https://github.com/bytecodealliance/wit-bindgen/issues/526). This is blocked until a resolution can be found to that issue.

philpax:
Future work: add a way to send data from one side to another, both reliable and unreliable, and expose synced_resources

PR #202, opened 2023-03-03
Make the Github Action work only when needed

C-BJ:


FredrikNoren:
@C-BJ Yeah I can see, looks good. Thanks for the patch!

PR #210, opened 2023-03-04
Suffix versions

### LINKED ISSUE #207: Support version suffixes in Ambient project version parser
In the `ambient_project` crate, we parse the `version` field in project manifests. However, I purposely kept it simple and only implemented `major.minor.patch` versioning, which means you can't add suffixes to the version like `-dev` or `-rc`. It would be nice to fix this so that we could automatically synchronise the project versions with that of the host, even when the host is using `-dev`.

The code for it can be found here:
https://github.com/AmbientRun/Ambient/blob/775df71875e96d7ced120648a3034ffe5f9301ff/crates/project/src/lib.rs#L243-L322

If you tackle this, make sure to add tests for ordered comparison (i.e.  `1.0` > `0.2.0-rc2` > `0.2.0-rc1` == `0.2.0-rc1` > `0.2.0-dev` > `0.1.12` > `0.1.0`) and round-trip serialization (i.e. `let version_roundtrip: Version = serde_json::from_str(&serde_json::to_string(version)); assert_eq!(version, version_roundtrip);`) work correctly.

gautamprikshit1:
Fixes #207

Need help with tests.


philpax:
I've added a test for version parsing, fixed the existing tests, and added a test for roundtrip serialisation. I hope that's sufficient to see how to do the ordered comparison - I'd take the roundtrip serialisation test, come up with a list of versions that are in sequence, and then confirm that each one is less than the one that came before (you should be able to do that with [slice::windows](https://doc.rust-lang.org/std/primitive.slice.html#method.windows)).

To test, run 

```sh
cargo test -p ambient_project
```

philpax:
Taking over this PR on Monday.

PR #211, opened 2023-03-04
feat: Entity::assert_all() (resolves #200)

### LINKED ISSUE #200: Add Entity attribute assertions
It would be useful to assert that all components in an `Entity` have a specific attribute, for example `Networked` to ensure that all components you are adding to the entity will also be mirrored on the client.

Currently, non-networked components are silently ignored, and can lead to bugs if you forget to annotate your components.

For example;
```rust
let entity = Entity::new().set(my_non_networked_component(), 5.);
entity.assert_all(Networked);
```

MavethGH:
This adds a function for asserting that all components of an `Entity` have a given attribute, panicking if any do not. Suggested by @ten3roberts in #200.

### DIFF
--- a/crates/ecs/src/entity.rs
+++ b/crates/ecs/src/entity.rs
@@ -177,8 +177,10 @@ impl Entity {
     /// Example:
     ///
     /// ```should_panic
-    /// # components!("doctest, {my_non_networked_component: String}")
-    /// let entity = Entity::new().set(my_non_networked_component(), "Can't serialize me!");
+    /// # use ambient_ecs::{components, Entity, Networked};
+    /// # components!("doctest", {my_non_networked_component: String,});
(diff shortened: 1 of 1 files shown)


MavethGH:
Note that I also ran rustfmt on `entity.rs`. I don't think that it'll be a problem, but it's worth mentioning since that's why git is saying that some of the imports changed.

MavethGH:
Also, I just noticed that the doctest doesn't compile :woman_facepalming:.

MavethGH:
> Also, I just noticed that the doctest doesn't compile woman_facepalming.

Fixed, after probably too much effort...

ten3roberts:
Looks fantastic.

Yeah, not sure about the imports as things *should* already be formatted but some mysteries are best left alone.

PR #212, opened 2023-03-05
Allow the use of namespaces in concept definitions

### LINKED ISSUE #108: Allow the use of namespaces in concept definitions
At present, concepts are restricted to single identifiers. This means that there's no way to organise a group of related concepts together, other than putting them in another project.

This should be relatively simple to do (switch to using `IdentifierPathBuf` instead of `Identifier` and update the codegen appropriately).

ArberSephirotheca:
resolve #108 

guest/rust/api/api_macros/src/ambient_project/ChangeLog:
```
* concept.rs: add tree structure for concept to allow the use of namespace
* manifest.rs: add concept to NamespaceOrOther enum
   change concepts field in Manifest to: BTreeMap<IdentifierPathBuf, NamespaceOrOther>
* component.rs: update and fix compiler error.
* test.rs: add test for the use of namespace in concept definition.
* mod.rs: change the body of implementation function for concept.
```
guest/rust/examples/basics/third_person_camera/src/ChangeLog:
```
* lib.rs: fmt
```

Tried to create generic tree structure that is capable of constructing and tokenizing both concept tree and component tree. However, that needs to change more codes and add few more abstractions, which I think is overcomplicated. So instead I create a tree structure specifically for Concept, and it is straightforward. but the drawback is we have some repetitive codes.


philpax:
Wow, nice work! I'll have a play around with this but this looks like it should solve the problem handily.

philpax:
So the three changes I made are:

- Testing on the host for namespaced concepts (I forgot to mention this, all good)
- Supporting namespaced extends in the guest code (e.g. `extends: Vec<IdentifierPathBuf>` instead of `Vec<Identifier>`)
- Unifying the two tree types: this wasn't actually too bad, you already did most of the work 😄 

Thanks for the great contribution, will merge when tests pass!

Issue #193, opened 2023-03-01
Add "Dump UI" button to the debug menu


philpax:
When you're in an Ambient game with a client, there are three ECS worlds: the UI world, the client world, and the server world. The debugger offers buttons to dump the Client World and the Server World, but does not offer a button to dump the UI world. This is something we accidentally left out.

If you'd like to get your feet wet with our UI system and/or get an understanding of how the  before it lands in the guest API, you can have a go at trying to add the button itself. The main thing you need to do is to clone this:

https://github.com/AmbientRun/Ambient/blob/41a6f86c3df12707a02a72b6f4c660fc63cfe6c0/crates/debugger/src/lib.rs#L56-L65

and dump the world passed in in the `Button`, and not in `get_state` (which gets the client world).

Things to watch out for:

- The hotkeys. They're already wrong (Show entities and DCW use the same hotkeys) - feel free to go ahead and reassign them.
- `dump_world_hierarchy_to_tmp_file` doesn't take a path. You should update this.
  - If you want to take this further: update *all* of the dump methods to use a unified code path for writing files, so that they all behave consistently w/r/t logging and output location. This may also be a good opportunity to add a button that dumps all of the state at the same time for easy debugging.

Feel free to ask any questions!

owenpalmer:
I'll give it a go!

owenpalmer:
@philpax Okay I got the new button dumping the UI world state.

1. As for the custom file path on dump_world_hierarchy_to_tmp_file, should all world_dumps go to the tmp directory, but allow for subdirectories within it?
2. When you talk about a unified code path for writing files, is this referring to all dumps or just world dumps?
3.For the hotkeys, I was thinking they could be from F1-F8, starting in order from "Show Entities" on the right to "Shader Debug" on the left. Thoughts?

philpax:
> @philpax Okay I got the new button dumping the UI world state.
> 
>     1. As for the custom file path on dump_world_hierarchy_to_tmp_file, should all world_dumps go to the tmp directory, but allow for subdirectories within it?
> 
>     2. You mentioned a unified code path for writing to files, is this referring to all dumps or just world dumps?
> 
>     3. For the hotkeys, I was thinking they could be from F1-F8, starting in order from "Show Entities" on the right to "Shader Debug" on the left. Thoughts?

[...]

PR #216, opened 2023-03-09
Add Dump UI World button and remap hotkeys

owenpalmer:
Fixes https://github.com/AmbientRun/Ambient/issues/193.

There was a bug that prevented `hotkey_modifier` from working, the modifiers weren't being set in the keyboard event.
I decided not to implement the write_debug function in this commit since I would rather not involve it in a minor feature such as this.

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -9,7 +9,6 @@ use ambient_core::{
     camera::camera_systems,
     frame_index,
     gpu_ecs::{gpu_world, GpuWorld, GpuWorldSyncEvent, GpuWorldUpdate},
-    hierarchy::dump_world_hierarchy_to_tmp_file,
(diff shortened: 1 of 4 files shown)


crates/app/src/lib.rs
```diff
@@ -610,9 +609,7 @@ impl System<Event<'static, ()>> for ExamplesSystem {
                     },
                 ..
             } => match virtual_keycode {
```
FredrikNoren, line 611: So these are actually used in the "native" examples, which doesn't have a Debugger. I'll merge and just revert that part

owenpalmer: Okay that makes a lot more sense, thank you!

FredrikNoren:
@owenpalmer Merged. Thanks for the PR!

owenpalmer:
Happy to help!

Issue #213, opened 2023-03-06
Physics API improvements


philpax:
As requested by @SK83RJOSH and with other suggestions by me. I'm publicly sketching these out before I implement them:

- #162 
- #161 
- #182 
- Reading / setting `mass` through the component
- `physics::get_velocity_at_position`: a way to read the velocity of a physics object at a given position relative to the object. Need to make it clear what coordinate space it's in.
- `physics::apply_force` should have `ForceMode` parameter (Acceleration, Force, Impulse, VelocityChange) and (maybe) control over whether the object should be woken
  - `physics::apply_force_at_position` or an optional `position` parameter for `apply_force`. Need to make it clear what coordinate space it's in.
-  `physics::toggle_collisions_between(ent1, ent2, enabled)`: there's quite [a bit of nuance](https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/RigidBodyCollision.html#collision-filtering) here. Not sure if this is the API we should expose, or if we should expose collision layers, or if users should be able to register their own collision handlers. Need to study what other engines do here.
- `physics::sweep`: like `raycast`, but tells you [if an object will collide with something else](https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/SceneQueries.html#sweeps). 
- `sweep` and `raycast`: Expose *all* available information. At present, we return the collided-with entity and the distance at which the collision should occur. PhysX also provides us with the exact position, normal, and information about the face index/terrain UV. The normal especially is useful for gameplay.
- `inertia_tensor: Vec3` component, like [Unity's `inertiaTensor`](https://docs.unity3d.com/ScriptReference/Rigidbody-inertiaTensor.html).

Something else to keep in mind that whatever we expose should be portable to other physics engines, too. We may support other physics engines to enable certain applications / deployments.

Feel free to suggest other features of interest that you would like exposed. I'd like to ensure that they all have a consistent interface, which is why I'm not going to nail any of these down until I have a good idea of what should be implemented.

SK83RJOSH:
For collision filtering, we're going to need to make some refactors to the physics set up in the ECS:
- [ ] Remove `physics_shape(): PxShape`, and instead create a new entity for each shape
- [ ] Add a new component to each physics entity `physics_shapes(): Vec<EntityId>`
- [ ] Add a new component to each physics shape `physics_filter_data(): PxFilterData`
- [ ] Devise a solution for physics layers
[...]

PR #235, opened 2023-03-19
Physics API improvements

SK83RJOSH:
Address a few of the tasks in #213


philpax:
Looks good to me. Is this ready to merge, aside from #236?

SK83RJOSH:
Negatory, half the force modes don't work for at position; going to refactor this into separate functions for force/impulse and ignore the other force modes until we've exposed mass in a nice way (since it's required for the force modes ignoring mass)

PR #254, opened 2023-03-24
Update quinn to 0.9

pierd:
Newer version has a bit nicer API.

It uses custom futures instead of stream so in many cases one can get away with immutable borrows (stream needs mut) and there's no need for extra `Option` that comes from using stream's `next()` directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5221,16 +5221,16 @@ checksum = "658fa1faf7a4cc5f057c9ee5ef560f717ad9d8dc66d975267f709624d6e1ab88"
 
 [[package]]
 name = "quinn"
-version = "0.8.5"
+version = "0.9.3"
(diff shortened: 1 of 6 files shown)


FredrikNoren (approved):

ten3roberts:
Just a note on this:

For web support we will depend on h3 and h3_quinn, which is still on `0.8.3` so we will get two different versions of quinn in our dependency tree for our network stack

pierd:
> Just a note on this:
> 
> For web support we will depend on h3 and h3_quinn, which is still on `0.8.3` so we will get two different versions of quinn in our dependency tree for our network stack

Good catch. I've noticed that there's already an issue to update `quinn` dependency in `h3_quinn` to 0.9 (https://github.com/hyperium/h3/issues/144).

PR #259, opened 2023-03-28
WIP: Remove server input

### LINKED ISSUE #252: Remove input API from server
With #240, it is now possible for developers to send their own custom messages to the server. We should avoid baking in any assumptions about input, so clientside WASM should be responsible for picking up relevant inputs and sending them to the server.

philpax:
Closes #252.

This PR removes all sending of player input and window sizes to the server, and uses messages to send it from client WASM instead. This has some up- and downsides, but I think it works out in the long run (especially once we have dependencies for implementing common operations).

- [x] Port tictactoe
- [x] Port minigolf
- [x] Port pong


philpax:
CI not responding, closing and reopening

philpax:
Need to provide 
- function to get user id from source if available
- faster user id to entityid lookup

PR #260, opened 2023-03-29
Use ResultEmpty and OkEmpty in new_project_template

owenpalmer:
Looks like this was missed in https://github.com/AmbientRun/Ambient/commit/abe3e624c12cc463da2862d076070eedac249421


philpax:
Good catch! That's what I get for only running the rename in the guest workspace 😅

owenpalmer:
Thx! Happy to help :)

PR #275, opened 2023-03-31
Add proxy

pierd:
Addresses: https://github.com/AmbientRun/InternalIssues/issues/2

The actual proxy is implemented in https://github.com/AmbientRun/AmbientProxy. The repository also contains the client part (which in this case is the ambient server) so an integral part of this PR is: https://github.com/AmbientRun/AmbientProxy/tree/main/src/client.rs

The proxy is transparent to the client part (`ambient join`) so the joining users only need to know the endpoint to connect to but otherwise they don't have to know that the connection goes via proxy.


app/src/cli/mod.rs
```diff
@@ -101,6 +101,10 @@ pub struct HostCli {
     /// Defaults to 9000
     #[arg(long)]
     pub quic_interface_port: Option<u16>,
+
+    /// AmbientProxy address to use for NAT traversal
+    #[arg(long)]
+    pub proxy: Option<String>,
```
FredrikNoren, line 107: Can we make it so it defaults to proxying? I.e. when you run `ambient serve` or `ambient run` it just prints something like "use `ambient join proxy.ambient.run:324` to join". We can add a `--no-proxy` flag to turn it off.

I know this might be a bit "wasteful" in the future if there's a lot of local development and we can see about revisiting it then, but I just want to make it as easy as possible for people to create multiplayer experiences now. (Though we can probably hole-punch in the future also, right?)

crates/network/Cargo.toml
```diff
@@ -18,6 +18,7 @@ ambient_ui = { path = "../ui" }
 ambient_renderer = { path = "../renderer" }
 ambient_element = { path = "../../shared_crates/element" }
 ambient_app = { path = "../app" }
+ambient_proxy = { git = "https://github.com/AmbientRun/AmbientProxy.git", rev = "3944422629e0eeacde62080fbc5055593608395c" }
```
FredrikNoren, line 21: We should just publish the proxy to crates.io

FredrikNoren (approved):

FredrikNoren:
Might be worth lowering the semver bounds on the proxy crate so more stuff can be shared if the versions diverge?

app/src/cli/mod.rs
philpax: fwiw Gradio uses `--share` for this; applications are local by default, but you're given a proxy URL when you specify `--share` (no arguments). 

That seems like a reasonable path to me - I think people would prefer development to be local by default (I've definitely had my fair share of people joining my development servers when I didn't expect it!)

philpax: Yeah, fair enough, let's see how it goes 👍

philpax (approved):
LGTM. Will the changes to `Connection` cause any problems for you, @ten3roberts ?

pierd:
FYI Right now it will block the initialisation flow when waiting for connection to proxy. Without internet connection that's a 60s wait for timeout.

I'm moving the code a little bit to run the proxy connection in a separate task.

ten3roberts:
> LGTM. Will the changes to `Connection` cause any problems for you, @ten3roberts ?

Does not look like it, infact, the Connection *trait* makes my implementation for separate backends (quinn, webtransport) even easier.

Great work Kuba

PR #276, opened 2023-03-31
Run the Ambient project macro on the host

### LINKED ISSUE #228: Structured events
Right now events are just strings with an `Entity` as even data, but it would be nice to be able to have structured/typed events. Something like this in the `ambient.toml`:

```toml
[types]
DamageType = { enum = ["Fire", "Ice"] }

[events]
gun_fired = { damage = "u32", damage_type = "DamageType" }
```

Which could then be consumed like this:
```rust
on(gun_fired, |data| {
  println!("gun fired with {} damage", data.damage);
});
```

philpax:
This moves the macro implementation to `shared_crates`, and wires it up on the host and the guest. This should enable a lot of neat functionality, like components defined in one place (root `ambient.toml`) once that's implemented.

Closes #228.


philpax:
I think generating and evaluating `wit` would be a little difficult to do, but it might not be necessary - my implementation here allows for arbitrary ser/de of types as long as they're agreed on by both sides. Something like what you're proposing should be possible in that case!

Right now I'm working on converting events over to messages from the runtime, then I'll land this PR and think about structs and enums 🚀

Issue #258, opened 2023-03-28
Add clientside support for `asset::url`


philpax:
At present, `asset::url` is not supported for clientside WASM. This is because the current implementation is dependent on the server's base URL.

We should build a way to get the corresponding asset URL for a base path when on the client. It would be nice if this was designed in such a way to enable multiple projects of assets in the future, too.

philpax:
This has been done, but @pierd is currently working on improving support for arbitrary locations in his branch.

PR #286, opened 2023-04-05
support asset url in client

chaosprint:
related to https://github.com/AmbientRun/Ambient/issues/258

I change the `server_asset` to `asset` so both the client and server can call `asset::url`.

I added a `client.rs` in the `asset_loading` example for testing.


crates/wasm/src/shared/implementation/asset.rs
```diff
@@ -0,0 +1,20 @@
...
+};
+
+use ambient_ecs::World;
+
+pub(crate) fn url(
+    world: &mut World,
+    path: String,
+) -> anyhow::Result<Option<String>> {
+    let assets = world.resource(asset_cache());
+    let base_url = ProxyBaseUrlKey.try_get(assets).unwrap_or_else(|| ServerBaseUrlKey.get(assets));
```
FredrikNoren, line 17: ProxyBaseUrlKey and ServerBaseUrlKey won't exist on the client side when users join with `ambient join`. They need to be communicated to the client in that case (that's what the `synced_resource` thing I mentioned could be used for).

crates/network/src/lib.rs
```diff
@@ -40,6 +40,13 @@ pub const PLAYER_INPUT_DATAGRAM_ID: u32 = 5;
 pub const WASM_DATAGRAM_ID: u32 = 6;
 
 components!("network", {
+    /// The absolute asset url, synced as a component between client and server for loading assets.
+    @[
+        Debuggable, Networked, Resource,
+        Name["Absolute Asset Url"],
+        Description["URL that is absolute, and can be used to load assets from the client or server."]
+    ]
+    abs_asset_url: String,
```
FredrikNoren, line 49: Hm so the naming here feels weird; `AbsAssetUrl` is already a struct which is very different from this; the `AbsAssetUrl` struct is just a wrapper for a Url which point to an asset anywhere on the internet (and it can't be a relative path).

Something like `content_base_url` would make more sense to me.

app/src/server/mod.rs
```diff
@@ -89,6 +91,7 @@ pub fn start(
         let name = manifest.project.name.clone().unwrap_or_else(|| "Ambient".into());
         server_world.add_components(server_world.resource_entity(), Entity::new().with(project_name(), name)).unwrap();
 
+        Entity::new().with(synced_resources(), ()).with(abs_asset_url(), format!("http://{public_host}:{http_interface_port}/content/")).spawn(&mut server_world);
```
FredrikNoren, line 94: Hm so the `ProxyBaseUrlKey` seems to be lost now; the asset::url method used to use both but now it's just using ServerBaseUrl

FredrikNoren (approved):

Issue #302, opened 2023-04-13
Update `wasmtime_wasi` version


philpax:
Right now, we have a version of `preview2-prototyping` that I brought into tree manually and converted to sync from async. It would be nice to do this a little more automatically so that we can track changes more generally.

I believe updating should also help address the issues we're seeing with *all* script output being routed to stderr, which is non-ideal.

PR #303, opened 2023-04-13
Campfire: a tool for project-specific tasks

### LINKED ISSUE #146: Automate the release process
At present, our release process requires a few manual steps, as seen in [contributing](https://github.com/AmbientRun/Ambient/blob/main/docs/src/contributing.md). We'd appreciate some help with automating this.

We had a POSIX-only [release script](https://github.com/AmbientRun/Ambient/blob/453dc906238ee9cd2cb4e1218acac5a829e7eaf7/scripts/release.sh) using the individual steps of `cargo release`, but we stopped using it because it required SSH authentication on each step and each step took several seconds to run. We have unique needs because of our two workspaces (the runtime and the Rust guest), so we can't use `cargo release` as is.

Our CI takes care of publishing, so all we need is a cross-platform way of running through the steps outlined in `contributing.md` automatically.

philpax:
There are a few tasks that come up often enough in Ambient development that it's worth automating them. Campfire is an in-repo tool for these tasks.

Closes #146.

- [x] `campfire doc`: generates documentation from rustdoc output. Replaces the `generate-docs` example.
- [x] `campfire example clean`: removes all build artifacts for all examples
- [x] `campfire example run [name]`: runs the example `name`; both `basics/skinmesh` and `skinmesh` work
- [x] `campfire example run-all`: replaces `scripts/run_all_examples`; runs all examples in sequence
- [x] `campfire example check-all`: checks all examples for compilation/Clippy warnings
- [x] `campfire release update-version x.y.z`: updates versions in all relevant files to point to `x.y.z`
- [x] `campfire release update-msrv x.y.z`: updates all MSRVs for all packages, including the Dockerfile
- [ ] `campfire release check`: checks the following:
  - the `Dockerfile` can run an Ambient server
  - the MSRV is correct for both the host and the API
  - both the runtime and the guest can build with no errors
  - the CHANGELOG's unreleased section is empty
  - `README.md` and `docs/src/introduction.md` match their introductory text

Maybe:
~~- [ ] `campfire wasi update` applies patches to a freshly-cloned version of `wasmtime_wasi` from `preview2-prototyping` (#302)~~


FredrikNoren:
Nice! We should probably throw the golden image tests in there at some point too

PR #317, opened 2023-04-16
Rust-native audio transcoding

### LINKED ISSUE #38: Use Rust-native code to transcode audio in the audio pipeline
At present, we shell out to `ffmpeg` to transcode audio to `ogg`. To reduce the dependency on external utilities (especially on Windows), we should consider using the `symphonia` crate to decode all supported audio formats, and then encode it with one of the `ogg` encoding crates.

marceline-cramer:
Closes #38.

Regrettably, there are no pure-Rust ogg Vorbis encoding crates right now. However, I've added the next best thing: [vorbis_rs](https://crates.io/crates/vorbis_rs). It has active maintainers and uses the latest ogg libraries. The other Rust ogg libraries are unmaintained and use older versions of libogg and libvorbis with known security vulnerabilities.

I also caught an issue where sound graph files were using the string literal `"SOUND_GRAPH_EXTENSION"` instead of the constant value `SOUND_GRAPH_EXTENSION`, so I've patched that.

Also, I haven't implemented the copying of tags from the input file to the output file. The reason is that `vorbis_rs` expects tag values to be strings but symphonia tag values can be [lots of different types](https://docs.rs/symphonia-core/latest/symphonia_core/meta/enum.Value.html). I'm not sure what to do about this, so I've left tags empty in the output file for now.


chaosprint:
Just played with the `physics` examples using a `wav` file and got this error:

```
[2023-04-17T07:13:58Z INFO  symphonia_core::probe] found the format marker [52, 49, 46, 46] @ 0+2 bytes.
[2023-04-17T07:13:58Z ERROR ambient_build] In pipeline pipeline.json/1, at file ping.wav
    
    Caused by:
        0: Failed to create audio decoder
        1: unsupported feature: core (codec):unsupported codec
```

chaosprint:
Hi,

It looks that `Symphonia` still has some work to do. So we have to wait or switch to other solutions.

chaosprint:
> 

For extra featues, probably not. But I do have `minimp3` and `hound` working on my Mac while for `symphonia`, I cannot run its official examples. Have you tried to run the `symphonia` examples?

marceline-cramer:
Yeah, I've ran `symphonia-play` from the latest `master` branch and it works fine on every audio file I've given it so far. For extra context I'm on Artix Linux. Do you know what's keeping you from running the examples?

marceline-cramer:
I made a mistake with the channel count; counting the bits WAS the correct way to do that, but I confused the `.bits()` function (which returns the numeric bitfield) with the `.count()` function (which actually counts the bits).

After enabling the `pcm` Symphonia feature and counting the bits correctly, the stereo audio files I've tested now get converted to `ogg` with no issue.

chaosprint:
> I made a mistake with the channel count; counting the bits WAS the correct way to do that, but I confused the `.bits()` function (which returns the numeric bitfield) with the `.count()` function (which actually counts the bits).
> 
> After enabling the `pcm` Symphonia feature and counting the bits correctly, the stereo `.wav` files I've tested now get converted to `ogg` with no issue.

Cool, I will test this ASAP

philpax:
Sounds good to me - the log hiding is done at the top-level here:

https://github.com/AmbientRun/Ambient/blob/40eb6eb6449f48e2b2c10b7c74fdb788f260958f/app/src/main.rs#L28-L40

We can just add a commit to this PR to do that, and merge (Marceline won't be online until later) - are you OK to do that?

chaosprint (commented):
The latest version works perfectly on my Mac. During the conversion, some unnecessary logs are printed, but I guess it's harder to remove as this comes from the symphonia itself. We can work on the conversion mechanism later: https://github.com/AmbientRun/Ambient/pull/320
```
[2023-04-19T08:21:42Z INFO  symphonia_core::probe] found the format marker [49, 44, 33] @ 0+2 bytes.
[2023-04-19T08:21:42Z INFO  symphonia_core::probe] found the format marker [ff, fb] @ 1246+2 bytes.
[2023-04-19T08:21:42Z INFO  symphonia_bundle_mp3::demuxer] using xing header for duration```

Issue #136, opened 2023-02-21
Capture mouse / hide cursor


FredrikNoren:
We need some way to capture the mouse/hide the cursor. Probably dependent on #6

philpax:
What should the API for this look like, and what should it allow you to do?

kevzettler:
I would suggest copying the high level patterns of the browsers Pointer Lock API:
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API

This may have the added benefit for the Ambient browser target because you will need to use that anyways for browser builds.

PR #319, opened 2023-04-17
rotate the camera without right click

chaosprint:
It would be more intuitive if the camera can rotate by mouse moving immediately since there is no text description. When I approached this demo, I had some difficulties figuring out how to rotate the camera. `MouseButton::Right` doesn't seem to work for me as I use BetterTouchTool's two-finger tap as a right click and disable Mac Trackpad's secondary click in Mac preferences. May need to open another issue if anyone else reports.


philpax:
My reason for adding this initially was because the no-right-click behaviour was very annoying when you hit the borders of the window, or switched windows. I'm working on #136 and that should hopefully let us fix this properly.

philpax:
This now works as expected - the camera will rotate, and you can hit Esc to release the input lock

PR #320, opened 2023-04-17
load mp3/wav file with its own extension

chaosprint:
Even if in the future we switch ffmpeg to pure rust solutions, we may still need to avoid inexplicit extension change from `wav` or `mp3` to `ogg`. The logic is that if we do the format conversion secretly, we should do the URL conversion secretly as well to avoid extra cognitive load for users/developers.


philpax:
I agree with the change (it *is* confusing to change the extension), but I'm not sure if replacing the extension in `asset::load` is the best place to do it. 

Maybe we should have the build system create a folder and keep the file in there, so that you can pass in `assets/shoot.wav` to `asset::url`, which will look up `ambient-asset:/assets/shoot.wav/out.ogg`?

That would also allow us to have different-quality sounds if need be.

chaosprint:
> I agree with the change (it _is_ confusing to change the extension), but I'm not sure if replacing the extension in `asset::load` is the best place to do it.
> 
> Maybe we should have the build system create a folder and keep the file in there, so that you can pass in `assets/shoot.wav` to `asset::url`, which will look up `ambient-asset:/assets/shoot.wav/out.ogg`?
> 
> That would also allow us to have different-quality sounds if need be.

Good idea. I will work on the audio loading part to support different formats.

chaosprint:
replaced by https://github.com/AmbientRun/Ambient/pull/344

Issue #283, opened 2023-04-03
Add helpers for 3D scene querying


philpax:
i.e. `world_to_screen`, `screen_to_world_direction`, `screen_ray`.

We already have `screen_ray`, so shouldn't be too hard to implement. Should we do this in the host or in the guest code?

owenpalmer:
Hey! I wouldn't mind giving this a shot, since it directly relates to my game, and I would love to learn more about how the api works!

philpax:
Awesome! I'm planning on writing up some documentation about how the WIT stuff works, but in the meantime, you can poke around the commits for the `wit` folder: https://github.com/AmbientRun/Ambient/commits/main/crates/wasm/wit

Basically, all of the core function definitions happen there; the host build script will automatically generate bindings on the host and guest. You'll then need to implement the corresponding traits on the host (just follow the errors) and write wrapper functions on the guest.

Let me know if you need a hand - I'm hoping to get more detailed docs written soon!

owenpalmer:
Thanks for the info @philpax! Just to get familiar with the process of exposing _**anything**_ to the api, I duplicated the [`get_raw_input`](https://github.com/AmbientRun/Ambient/blob/59aaf3cc1c99181f84b625c85a42d1e950ce05ee/crates/wasm/src/client/implementation.rs#L49) function (keeping the same fn body and return type, just to see how the process works), created a new wit definition, and ran the build script which successfully generated the bindings. I wrote the guest wrapper function, and successfully called the function from the API in Puckle! So I at least know how to expose a function :)

I have a few questions to find the right mental model of how this finalized implementation should work.

1. Would it make sense to expose the existing [`screen_ray`](https://github.com/AmbientRun/Ambient/blob/59aaf3cc1c99181f84b625c85a42d1e950ce05ee/crates/core/src/camera.rs#L302) function by simply calling it from the new host trait implementation (i.e if new `screen_ray` was implemented [here](https://github.com/AmbientRun/Ambient/blob/59aaf3cc1c99181f84b625c85a42d1e950ce05ee/crates/wasm/src/client/implementation.rs#L48))? Or should I completely move the function from `camera.rs` to the new trait implementation?
2. Does it make sense to expose `screen_ray` as a camera helper function similar to `get_raw_input`?
[...]

PR #331, opened 2023-04-19
Rename screen_ray to clip_space_ray

owenpalmer:
Fixes a task in this comment https://github.com/AmbientRun/Ambient/issues/283#issuecomment-1513477893


owenpalmer:
@philpax Looks like they failed due to a problem with the action. Same thing happened here https://github.com/AmbientRun/Ambient/pull/329

philpax:
Awesome! Do you think you'd be able to add some text on top of where your cursor is to the `screen_ray` example? Should be able to create a `Text.el()` that updates its `translation` each frame (using the `use_frame` hook)

owenpalmer:
What's the best way to mutate the text Element's translation component from inside `use_frame`? I'm currently a little unfamiliar with the UI API. Thx :)

philpax:
Hm, just realised that's not a simple suggestion because you'll need a way to get the world position back from the server. OK, ignore that for now and I'll add it once I do some other work. Is the PR good to go otherwise?

philpax:
Ended up doing the UI stuff because I realised we didn't have `use_module_message`. If you're okay with everything, I'll merge 🚀

owenpalmer:
Cool, yep everything looks good to me! Thanks for doing that UI stuff :)

PR #332, opened 2023-04-19
improve audio API

chaosprint:


crates/audio/src/source/mod.rs
```diff
@@ -146,7 +154,7 @@ pub trait Source: Send {
         SampleIter::new(self)
     }
 
-    fn gain(self, gain: f32) -> Gain<Self>
+    fn gain(self, gain: Arc<Mutex<f32>>) -> Gain<Self>
```
ten3roberts, line 157: Maybe we should make this generic to avoid breaking and making it impossible to use a plain literal gain

guest/rust/api/src/audio.rs
```diff
@@ -23,23 +23,42 @@ pub struct AudioTrack {
...
+        self
     }
 
     /// Set the volume of the track.
-    pub fn volume(&self, volume: f32) -> Self {
-        Self {
-            volume,
-            ..self.clone()
-        }
+    pub fn volume(&mut self, volume: f32) -> &mut Self {
```
ten3roberts, line 32: Should the volume be clamped from [0,1)?

In some cases you want to make the source *louder* than normal if it is very quiet, I've had to do that for previous effects.

Also, what is the difference between `gain` `volume` and `amp`?

chaosprint: > Should the volume be clamped from [0,1)?
> 
> In some cases you want to make the source _louder_ than normal if it is very quiet, I've had to do that for previous effects.
> 
> Also, what is the difference between `gain` `volume` and `amp`?

Good point. Need to unify the usage here. I actually had another term `scale` 😂

crates/audio/src/source/mod.rs
chaosprint: Fixed

guest/rust/api/src/audio.rs
chaosprint: I have removed the clamp. For the internal API, I suggest that we keep it as is. The users should only need to know and use `volume` at the moment.

chaosprint:
@philpax can you help with the merge?

app/src/client/wasm.rs
```diff
@@ -28,19 +28,57 @@ pub fn initialize(world: &mut World) -> anyhow::Result<()> {
...
+                AudioMessage::Track(t, looping, amp, url, uid) => {
+                    let gain = Arc::new(Mutex::new(amp));
+                    let gain_clone = gain.clone();
+
+                    let sound = match looping {
+                        true => stream.mixer().play(t.decode().repeat().gain(gain_clone)),
+                        false => stream.mixer().play(t.decode().gain(gain_clone)),
+                    };
+                    sound.wait();
+                    sound_info.insert(uid, (url, looping, gain, sound.id));
```
philpax, line 43: I'd suggest using a struct instead of a tuple here

crates/wasm/src/client/implementation/specific.rs
```diff
@@ -187,6 +192,45 @@ impl wit::client_audio::Host for Bindings {
...
+        runtime.spawn(async move {
+            async_run.run(move |world| {
+                let sender = world.resource(audio_sender());
+                sender.send(AudioMessage::Stop(url)).unwrap();
+            });
+        });
+        Ok(())
+    }
+
+    fn set_amp(&mut self, url: String, amp: f32) -> anyhow::Result<()> {
```
philpax, line 209: `set_volume` - even if the users don't see it, it'll be confusing for us + anyone else who ends up writing WIT bindings for their language

philpax (changes requested):
LGTM but there are a few things that I've suggested

crates/wasm/src/client/implementation/specific.rs
chaosprint: fixed

app/src/client/wasm.rs
chaosprint: fixed

chaosprint:
replace by https://github.com/AmbientRun/Ambient/pull/341

PR #335, opened 2023-04-19
Make sure that client bytecode url always uses slashes

pierd:


philpax (approved):
Seems reasonable to me, should this function be used anywhere else?

pierd:
> Seems reasonable to me, should this function be used anywhere else?

I tried looking for other places where we translate path directly to an URL but we seem to always have an URL coming from the scripts or parsed from project dir.

philpax:
Sounds good, hit it 🚀

PR #337, opened 2023-04-19
Document Element

philpax:
Partially complete, need to finish this


shared_crates/element/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+//! Element is a React-inspired virtual tree library for the Ambient runtime.
+//!
+//! It is backed by the Ambient ECS, which is used as the source of truth for the tree.
```
FredrikNoren, line 3: Hm the ECS isn't really the source of truth; it's rather the opposite that the UI state produces the entities. I.e. the data flows UI state -> ECS, not the other way around

philpax: Will clarify

PR #339, opened 2023-04-20
refactor(element): implicitly box return fns

philpax:
While doing #337, I noticed that the `Box`ing could be done automatically for the return functions in `use_spawn` and `use_effect`. Pretty straightforward ergonomics improvement.


FredrikNoren (approved):
Nice!

PR #341, opened 2023-04-20
Add music sequencer client-side and improve audio API

chaosprint:


philpax (changes requested):
LGTM. Just one last thing, can you add a `CREDITS.md` to `music_sequencer` with where the samples come from?

PR #343, opened 2023-04-21
WIP: Use results/errors for API returns

### LINKED ISSUE #236: Use results/errors for API returns
At present, all API functions are more or less infallible. Failures either panic or are implicitly ignored. We should instead have these return `result<_, domain-specific-error-variant>`.

philpax:
Kept putting this off, but it's time. Closes #236.

Will try to minimise breakage, but some will be inevitable.


philpax:
I'll do this bit by bit. Will merge this one in to begin with.

PR #344, opened 2023-04-21
Improve audio pipeline; audio format convert control

chaosprint:
Since the music sequencer uses some `wav` samples, I picked up the previous unfinished PR: https://github.com/AmbientRun/Ambient/pull/320

Now it's possible to control whether to convert `wav` or use it as is in the pipeline:

```
  {
      "pipeline": {
        "type": "Audio",
        "convert": false
      }
  }
```

I also documented the usage.

For `mp3`, I still feel that it's better to do the replacement in URL if the users still want to use `.mp3`, which is quite natural. There has to be a conversion anyway, so it's better to do the URL check/replacement in case of a 404 error.

`wav` is a different story: if converted, they must use `ogg`, and users should have known this if they specify it in the `pipeline.jons` by themselves.

### DIFF
--- a/crates/build/src/pipelines/audio/mod.rs
+++ b/crates/build/src/pipelines/audio/mod.rs
@@ -1,6 +1,7 @@
 use ambient_std::asset_url::AssetType;
 use ambient_world_audio::AudioNode;
 use anyhow::Context;
+use serde::{Deserialize, Serialize};
 use tracing::{info_span, Instrument};
 
 use super::{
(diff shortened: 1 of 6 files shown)


philpax (approved):
This is fine for now, but I'm not a huge fan of preprocessing the URL based on the filetype (it'll lead to really inconsistent behaviour).

I've been thinking about having the build process generate a manifest that can be used to map input files to output files for this purpose, and more - I'll write up an issue for that.

Issue #292, opened 2023-04-09
Add decals example


FredrikNoren:
There's already https://github.com/AmbientRun/Ambient/blob/main/crates/decals/examples/decals.rs, we should just move that to the guest examples

PR #347, opened 2023-04-24
Add Decals example

kevzettler:
Hello I need feedback and code review on this PR. This addresses #292 I added a new `decal_from_url` component per @FredrikNoren suggestion at:
https://discord.com/channels/894505972289134632/1078283561540530216/1096581219925377195

The example seems to render okay:
image

I suspect this can be cleaned up or migrated. I left the old `decal` component in the code because there were other crates that were dependent on it like the prefab. Not sure how to decouple or migrate the other crates. Please offer some guidance.


guest/rust/examples/basics/decals/src/server.rs
```diff
@@ -0,0 +1,57 @@
...
+        .with(color(), vec4(0.5, 0.5, 0.5, 1.))
+        .with_default(cast_shadows())
+        .spawn();
+
+    Entity::new()
+        .with_default(quad())
+        .with(scale(), Vec3::ONE * 10.)
+        .spawn();
+
+    Entity::new()
```
FredrikNoren, line 32: I think since we're anyway adding the cube stuff in the system above, you might not need the `.with_default(cube())` here?

crates/decals/assets/checkerboard.png
FredrikNoren, line 1: This file isn't used right?

FredrikNoren (approved):
Awesome work! Yeah leaving the `decal` component as is makes sense.

I can see that there's some issue with decal rendering in general though; it should only be visible inside of the green box. But I think that's a separate issue from just exposing the functionality to the guest side, so we can merge this and then create a separate issue for fixing decals in general. (I also see that the problem exists in the native decal example).

FredrikNoren:
Created an issue for the rendering issue here: #352

crates/decals/assets/checkerboard.png
kevzettler: This is used for the example at crates/decals/examples/decals.rs

guest/rust/examples/basics/decals/src/server.rs
kevzettler: removed in [b8ed8aa](https://github.com/AmbientRun/Ambient/pull/347/commits/b8ed8aa69d36ea3c6f01c24914faca4e91d5f91b)

PR #351, opened 2023-04-25
Music sequencer tweaks

philpax:
For the blog post, I'll likely be taking excerpts from the music sequencer to demonstrate certain concepts. I wanted to make sure that the code was as Ambient-idiomatic as possible.

I thought about it this morning and realised that each track could be treated as an ECS entity, which simplifies the logic considerably (no more huge-array-of-notes) and allows for other fun things, like dynamically adding tracks.

I've also added a BPM slider to demonstrate that nothing's static, aside from the number of notes.


chaosprint (approved):
I have tested it a little bit and it works well. BPM control is great. BTW, I had this idea a few days ago to modify the sequencer into like a demo of Firefly sync algorithm, which could also show off the ability for clients to send messages to each other. But for current version, BPM control would be enough I guess.

philpax:
> Any chance we can make the example any shorter though? BPM selection is nice, but the quicker people can understand the whole thing the better, so might be worth having fewer features for that.

I thought about that, but I think this is a pretty good all-encompassing demonstration of how a networked UI application might be structured.

What I might do for the blog post is to demonstrate a simplified version without BPM control, per-player colours, and maybe swapping out the HSV code for `palette`.

PR #359, opened 2023-04-28
Disconnect event loop from window

FredrikNoren:


FredrikNoren:
So the idea here was to support running multiple windows in a webpage, but it seems like even if we do this, the EventLoop is still consumed when it's started, so we can't create new windows after it's started anyway.

PR #362, opened 2023-05-03
Add(shared api): set_animation_blend, get_animation_clips, has_animation_clip

devjobe:
A bit of rushed API changes for manual animation blending and updated the skinmesh example to show usage. 

I don't think they are particularly good. I'd rather see a command based blending API but maybe it's good enough for 0.2.

* set_animation_blend: Shared api, sets weights and time, both as optional. Time is either absolute or percentage.
* has_animation_clip: Returns true when the asset cache has resolved an animation clip
* get_animation_clips: Get duration and binder ids. This one is questionable. I guess retargeting should be accounted for
* skinmesh example updates: Blending and time scrubbing


crates/wasm/wit/entity.wit
```diff
@@ -24,6 +24,22 @@ default interface entity {
         apply-base-pose: bool,
     }
 
+    /// Animation clip data
```
FredrikNoren, line 27: Hm might be a bit more rust-like to do `result<animation-clip, animation-error>`.
`enum animation-error { not-loaded, failed-loading(string) }` (not sure what the exact syntax is).

Also naming: `animation-clip` sounds like it's the actual clip. Maybe `animation-clip-metadata` or something (and same for `get-animations-clips`, since it's not returning the actual clip but just metadata).

crates/wasm/wit/entity.wit
```diff
@@ -24,6 +24,22 @@ default interface entity {
...
+        /// Duration in seconds
+        duration: float32,
+        /// If the asset is loaded (duration will be 0 and binders empty if false)
+        loaded: bool,
+        /// Error that occured while fetching asset
+        error: string,
+    }
+
+    set-animation-blend: func(entity: entity-id, weights: list<float32>, time: list<float32>, absolute-time: bool)
+    has-animation-clip: func(clip-url: string) -> bool
```
FredrikNoren, line 40: `has-animation-clip` sounds like it exists at all; maybe `animation-clip-loaded`?

FredrikNoren (approved):
Nice! Left some suggestions for improvements

crates/wasm/wit/entity.wit
devjobe: Added status and moved the api to asset.wit
```
    /// Cache status of an asset
    variant asset-cache-status {
        /// Ready to be used
        ready,
        /// Not yet loaded
        not-loaded,
        /// Asset has an error
        failed-loading(string)
    }
    /// Animation asset metdata
    record animation-asset-metadata {
        /// BinderId for the animation tracks
        binders: list<string>,
        /// Duration in seconds
        duration: float32,
        /// Cache status of an animation asset
        status: asset-cache-status,
    }

    get-animation-asset-status: func(clip-url: string) -> asset-cache-status
    get-animation-asset-metadata: func(clips: list<string>) -> list<animation-asset-metadata>
```

crates/wasm/wit/entity.wit
devjobe: Added status and moved the api to asset.wit

Issue #349, opened 2023-04-25
Build process should produce an output manifest


philpax:
At present, the only way to determine which output file corresponds to which input file is through trying to match up the built assets to the source files by folder hierarchy. This is confusing and often doesn't match intuition (a mesh produces a folder containing multiple folders containing multiple assets, while a WAV sound produces a single OGG file?)

To alleviate this and to simplify how assets are used in the API, I suggest producing an output manifest that lists all of the produced assets and provides a bidirectional mapping from source files to built assets.

This should help with the problem identified in #344, where sound files change format. Instead, users should be able to just specify the original source filename and have the runtime remap it to the actual output file based on the output manifest.

philpax:
The beginnings of this have begun with `metadata.toml` in the `build` directory, but it should cover more assets.

PR #363, opened 2023-05-03
Support running with URL as project path

pierd:
One of the requirements for [deployment](https://github.com/AmbientRun/InternalIssues/issues/15) is being able to run ambient project (`run`/`serve`) with an URL.


FredrikNoren (approved):
Ohhh, niice! Just skimmed the code but seems reasonable to me

philpax (approved):
Also skimmed over it, looks good! It looks like the `metadata` might also be the start of #349 ?

ten3roberts:
So if I've understood it this allows using an already running url as the "file server" for the project files, while the local command only does the computing?

Perhaps we should mention this in the book that it is possible to point the project to an url, similar to how vim allows ssh paths.

pierd:
> So if I've understood it this allows using an already running url as the "file server" for the project files, while the local command only does the computing?
> 
> Perhaps we should mention this in the book that it is possible to point the project to an url, similar to how vim allows ssh paths.

Yes, like a read-only file server for the built project files.

PR #364, opened 2023-05-03
feat(primitives): New capsule primitive

VinhTruongAmbient:
Exposes the internal `CapsuleMesh` to the guest API.

The commit also transforms the capsule geometry to match Ambient's coordinate
system. Capsule height is also changed to half-height in order to match PhysX's
definition.

Note: `uv_profile` is not exposed due to lack of `enum` support in `ambient.toml`.

Screenshot 2023-05-03 at 16 54 24


philpax (approved):
LGTM! Can you add a capsule to the `primitives` example?

PR #370, opened 2023-05-07
[win32 helper] add a bat file to help listing/building/running the examples on windows

mebyz:
**Goal :** 
Help windows users build and run the examples easily using their currently installed ambient runtime

**Implementation :**
A simple win32 bat file to be executed in the the [./guest/rust/]examples/ folder (this examples folder is also the one that is zipped upon all major releases - tags)

**Execution :** 
run_samples.bat will display (v0.2.0) :

run_samples_bat

### DIFF
--- a/guest/rust/examples/README.md
+++ b/guest/rust/examples/README.md
@@ -5,3 +5,7 @@ These are examples of Ambient projects written in the Rust programming language.
 Note that they must be run with a matching version of the runtime - if you're using a 0.1.1 runtime, you must use the 0.1.1 examples, and vice versa.
 
 You may encounter unexpected behaviour if you attempt to run the latest versions of the examples (i.e. `HEAD`/`main`) with an older runtime.
+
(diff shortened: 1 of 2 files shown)


chaosprint:
Hi,

Thanks for the work! Actually I have a similar PR before, adding an `.sh` file for UNIX system. But @philpax pointed out that we have already got https://github.com/AmbientRun/Ambient/tree/main/campfire
which provides an unified solution on different platforms like:
```
cargo campfire example run [EXAMPLE_NAME]
```
So, perhaps you can have a look at the folder and update the PR with the feature you want to add in `campfire`.

philpax:
Hm, I don't think this fills the same role as campfire - this should theoretically be included alongside the `examples.zip` to make it easier to run an example.

I'd like to avoid shipping batch files, but this serves a useful need in the meantime. Let's merge it for now 👍

PR #373, opened 2023-05-08
Fix block size

ten3roberts:


ten3roberts:
This branch is the result of me being bad at git and setting up my upstreams wrong, so that a new branch was pushed rather than a remote PR branch

PR #377, opened 2023-05-09
Add casts to physxx where c_char is involved

pierd:
Because `c_char` has different signedness on different targets (it's `u8` on aarch64 on linux) but `physx-sys` 0.8.2 uses `i8` in the generated code ([for example here](https://github.com/EmbarkStudios/physx-rs/blob/6b3a94534016270f8f6ebc665cd97ba7ce858d84/physx-sys/src/physx_generated.rs#L2069)).


philpax (approved):
what a headache

PR #379, opened 2023-05-10
feat(gpu): Simple graphics settings

VinhTruongAmbient:
Apps now try to find and load `settings.toml` from a specific directory. Currently the only two supported settings are the window size at boot time and VSync.

Present mode now prefers to VSync to avoid running apps at excessive frame rates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,6 +717,7 @@ dependencies = [
  "async-trait",
  "bytemuck",
  "byteorder",
+ "directories",
  "futures",
  "glam 0.22.0",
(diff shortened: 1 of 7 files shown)


philpax (approved):
Looks good to me. Do the examples that use the `app` crate also read the settings? It looks like it (and that's fine), just confirming.

VinhTruongAmbient:
> Looks good to me. Do the examples that use the `app` crate also read the settings? It looks like it (and that's fine), just confirming.

Yes they do.

PR #381, opened 2023-05-10
Network stack rework + Webtransport prep

ten3roberts:


crates/network/src/client_connection.rs
```diff
@@ -1,70 +1,75 @@
...
         }
     }
 
-    async fn send_datagram(&self, data: Bytes) -> Result<(), NetworkError> {
+    #[inline]
+    pub fn send_datagram(&self, data: Bytes) -> Result<(), NetworkError> {
         match self {
-            ClientConnection::Direct(conn) => Ok(conn.send_datagram(data)?),
-            ClientConnection::Proxied(conn) => Ok(conn.send_datagram(data).await?),
+            ConnectionKind::Direct(conn) => Ok(conn.send_datagram(data)?),
```
ten3roberts, line 70: This requires support in the proxy

crates/network/src/client_connection.rs
```diff
@@ -1,70 +1,75 @@
...
+            ConnectionKind::Proxied(conn) => todo!(),
+            // ConnectionInner::Proxied(conn) => Ok(conn.send_datagram(data).await?),
```
ten3roberts, line 72: I.e, sending datagrams to be sync

PR #383, opened 2023-05-11
Mesh position and index buffers are now required to exist

VinhTruongAmbient:
The rest of the code base asserts their existence so it makes sense to enforce it at import-time.

### DIFF
--- a/crates/model_import/src/gltf/mod.rs
+++ b/crates/model_import/src/gltf/mod.rs
@@ -15,6 +15,7 @@ use glam::{uvec4, Mat4, Quat, UVec4, Vec2, Vec3, Vec4, Vec4Swizzles};
 use gltf::animation::util::ReadOutputs;
 use itertools::Itertools;
 use relative_path::RelativePathBuf;
+use anyhow::Context;
 
 use self::gltf_import::GltfImport;
 use crate::{dotdot_path, model_crate::ModelCrate};
@@ -39,11 +40,9 @@ pub async fn import(import: &GltfImport, asset_crate: &mut ModelCrate) -> anyhow
             while let Some(tc) = reader.read_tex_coords(texcoords.len() as u32) {
                 texcoords.push(tc.into_f32().map(|x| x.into()).collect::<Vec<Vec2>>());
             }
-
-            let flip_indices = true;
             let mut cpu_mesh = Mesh {
                 name: format!("{}:{}:{}", import.name, mesh.index(), primitive.index()),
-                positions: reader.read_positions().map(|v| v.map(|x| x.into()).collect::<Vec<Vec3>>()),
+                positions: reader.read_positions().context("GLTF mesh must contain vertex positions")?.map(|a| Vec3::from(a)).collect::<Vec<Vec3>>(),
                 normals: reader.read_normals().map(|v| v.map(|x| x.into()).collect::<Vec<Vec3>>()),
                 tangents: reader.read_tangents().map(|v| v.map(|x| Vec4::from(x).xyz()).collect::<Vec<Vec3>>()),
                 texcoords,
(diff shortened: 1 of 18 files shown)


crates/model_import/src/gltf/mod.rs
```diff
@@ -52,22 +51,9 @@ pub async fn import(import: &GltfImport, asset_crate: &mut ModelCrate) -> anyhow
...
-                                let mut chunk = chunk.collect_vec();
-                                chunk.swap(1, 2);
-                                chunk
-                            })
-                            .collect::<Vec<u32>>()
-                    } else {
-                        v.into_u32().collect::<Vec<u32>>()
-                    }
-                }),
+                indices: reader.read_indices().context("GLTF mesh must contain an index buffer")?.into_u32().collect_vec(),
```
ten3roberts, line 54: Wont this unconditionally flip the indices/winding?

crates/std/src/uncategorized/mesh.rs
```diff
@@ -113,22 +101,18 @@ impl Mesh {
...
     }
     pub fn apply_skin(&mut self, joint_matrices: &[Mat4]) {
-        if let (Some(positions), Some(weights), Some(indices)) = (
-            &mut self.positions,
-            &self.joint_weights,
-            &self.joint_indices,
-        ) {
-            for ((position, weight), index) in
-                positions.iter_mut().zip(weights.iter()).zip(indices.iter())
+        if let (Some(weights), Some(indices)) = (&self.joint_weights, &self.joint_indices) {
```
ten3roberts, line 110: We could use `zip_eq` here if we expect them to be of the same length.

If not, we could use `izip!`to not have nested tuples

libs/unity_parser/src/asset.rs
```diff
@@ -33,8 +34,12 @@ impl Asset {
             BlendWeight,
             BlendIndices,
         }
-        let vertex_count = asset["Mesh"]["m_VertexData"]["m_VertexCount"].as_i64().unwrap() as usize;
-        let vertex_data_size = asset["Mesh"]["m_VertexData"]["m_DataSize"].as_i64().unwrap() as usize;
+        let vertex_count = asset["Mesh"]["m_VertexData"]["m_VertexCount"]
```
ten3roberts, line 37: Since we are now returning a result we could make the `[]` indexing fallible

ten3roberts:
I left some comments.

Apart from that it looks like a good change and is something that I started doing; namely making sure that if vertex uvs exist the position will exist as well.

Having `Option<Vec<_>>` is usally a tiny bit of code smell and implies there is a better solution since the `Vec` is already a functor, I.e; an empty Vec has the same meaning as None, and the difference between them is very subtle and if that difference is important it could be expressed in a different way

crates/model_import/src/gltf/mod.rs
VinhTruongAmbient: Yep, this was the case before as well, as `flip_indices` was a constant `true`.

ten3roberts: Oh I see now. Welp, if we need that behavior we have the history

crates/std/src/uncategorized/mesh.rs
VinhTruongAmbient: Yeah sure. I will have another look on a follow-up pass.

VinhTruongAmbient:
> I left some comments.
> 
> Apart from that it looks like a good change and is something that I started doing; namely making sure that if vertex uvs exist the position will exist as well.
> 
> Having `Option<Vec<_>>` is usally a tiny bit of code smell and implies there is a better solution since the `Vec` is already a functor, I.e; an empty Vec has the same meaning as None, and the difference between them is very subtle and if that difference is important it could be expressed in a different way

Thanks. I agree with making sure that the mesh is actually valid (all vertex attributes have equal number of elements etc). Might do it on a separate PR.

crates/model_import/src/gltf/mod.rs
VinhTruongAmbient: Yeah. If we wanted to do it properly, here's what the GLTF 2.0 spec says:

> When a mesh primitive uses any triangle-based topology (i.e., triangles, triangle strip, or triangle fan), the determinant of the node’s global transform defines the winding order of that primitive. If the determinant is a positive value, the winding order triangle faces is counterclockwise; in the opposite case, the winding order is clockwise.

PR #388, opened 2023-05-12
feat: non-subcommand specific flag support

ten3roberts:
This allows flags *before* subcommands like run which apply for every subcommand


ten3roberts:
This will be useful when adding things such as cert files or other "env" style configs

ten3roberts:
Damnit, messed up the cherry-picking

Will fix now

PR #389, opened 2023-05-12
wire up spatial audio

chaosprint:


crates/world_audio/src/systems.rs
```diff
@@ -10,10 +10,13 @@ use crate::{audio_emitter, audio_listener, audio_mixer, hrtf_lib};
 /// Initializes the HRTF sphere and adds the appropriate resources
 ///
 /// TODO: customizer IR sphere selection
-pub fn setup_audio(world: &mut World, mixer: AudioMixer) -> anyhow::Result<()> {
+pub fn setup_audio(world: &mut World, mixer: Option<AudioMixer>) -> anyhow::Result<()> {
```
ten3roberts, line 13: Why make it an option rather than using the mixer?

Another option would be to split it up so that the `setup_audio` isn't responsible for adding the mixer to the world as that is the only thing this function is using the mixer for.

ten3roberts (changes requested):

Issue #182, opened 2023-02-27
Support specifying custom physics materials


philpax:
At present, all physical colliders use a fixed `PxWoodMaterialKey`, which is a `PxMaterial` with `static_friction: 0.5, dynamic_friction: 0.5, restitution: 0.6`. This makes it quite difficult to control the physical behaviour of spawned objects, and is especially problematic for the minigolf example.

We'll probably want a way for the API to specify custom materials through assets?

droqen:
Reviving this. Would like to be able to create frictionless physics objects for reasons

PR #391, opened 2023-05-15
Simple procedural resource generation

### LINKED ISSUE #30: Basic procedural mesh / texture generation
This does not support updating meshes or more complicated functionality. This is purely to enable basic procedural mesh generation.

A basic API for creating assets at runtime, just to enable simple creation, and only on the client.

```rust
mod experimental {
  mod texture {
   fn create_2d(data: &[u8], format: texture::Format) -> Result<String, TextureCreateError> {}
  }

  mod material {
    fn create(diffuse_url: &str, /* material parameters */) -> Result<String, MaterialCreateError> {}
  }

  mod mesh {
    fn create(vertices: &[Vertex], indices: &[u32]) -> Result<String, MeshCreateError> {}
  }
}

use experimental::*;

let heightmap = make_heightmap();

let (vertices, indices) = convert_heightmap_to_mesh(&heightmap);
let mesh_url = mesh::create(&vertices, &indices)?;

let diffuse_url = convert_heightmap_to_diffuse(&heightmap);
let diffuse_tex_url = texture::create_2d(&diffuse_url, texture::Format::Rgba8Unorm)?;
let material_url = material::create(&diffuse_tex_url, /* ... */)?;

let id = make_transformable()
  .with(components::mesh_from_url(), mesh_url)
  .with(components::material_from_url(), material_url)
  .spawn();
```

VinhTruongAmbient:
Closes #30 

https://github.com/AmbientRun/Ambient/assets/129393025/c5a5004c-77a1-4f6b-98cd-8484992c0127

V1 Feature set:

- All APIs client-only! Creating procedural resources on `server.rs` is not supported.
- Procedural meshes
    - Position, texcoord, normal, tangent, must specify all
    - Indexed triangle lists only
- Procedural textures
    - Width, height, 2D only, no mipmaps
    - Supports almost all linear formats (no depth/stencil/compressed ones)
- Procedural samplers
    - Clamp to edge and repeat wrap modes
    - Nearest and linear filter modes
- Procedural materials
    - Pbr material only
    - Base color map (rgb + alpha), normal map, metallic roughness map
    - Transparency on/off

## Procedural generated mesh and material description

This mesh is a vertex and index buffer which are generated on the `guest` side.

procedural-material

## Procedural generated texture

This texture is generated with `noise` crate. 

Screenshot 2023-05-16 at 20 49 18

## Procedural generated sampler

Toggling between nearest and linear filters.

procedural-sampler


crates/renderer/src/lib.rs
```diff
@@ -166,6 +169,82 @@ pub fn systems() -> SystemGroup {
...
+                        todo!("handle error");
+                    };
+                    let Some(mesh) = mesh_key.try_get(&assets) else {
+                        todo!("handle error");
+                    };
+
+                    let dummy_material = FlatMaterialKey::white().get(&assets);
+                    let dummy_shader = cb(get_flat_shader);
+
+                    world.add_component(id, main_scene(), ()).unwrap();
```
philpax, line 204: Consider using `add_components` to do this all in one go

crates/wasm/wit/main.wit
```diff
@@ -13,6 +13,7 @@ world bindings {
     import client-input: pkg.client-input
     import client-camera: pkg.client-camera
     import client-window: pkg.client-window
+    import client-experimental: pkg.client-experimental
```
philpax, line 16: I'd suggest moving this to client-experimental-mesh; we're likely to have more experimental modules in the future

guest/rust/api_core/src/client/experimental/mesh.rs
```diff
@@ -0,0 +1,23 @@
...
+pub struct Vertex {
+    pub position: Vec3,
+    pub normal: Vec3,
+    pub tangent: Vec3,
+    pub texcoord0: Vec2,
+}
+
+pub fn create(vertices: &[Vertex], indices: &[u32]) -> String {
+    let vertices = unsafe {
+        std::slice::from_raw_parts(
```
philpax, line 15: I think we have bytemuck as a dependency? Consider using its `cast_slice` to ensure that size / alignment invariants are kept

guest/rust/examples/experimental/procedural_mesh/src/client.rs
```diff
@@ -0,0 +1,77 @@
...
+            let i2 = x + (y + 1) * resolution;
+            let i3 = (x + 1) + (y + 1) * resolution;
+            let mut v0 = temp_vertices[i0 as usize];
+            let mut v1 = temp_vertices[i1 as usize];
+            let mut v2 = temp_vertices[i2 as usize];
+            let mut v3 = temp_vertices[i3 as usize];
+            let n10 = (v1.position - v0.position).normalize();
+            let n20 = (v2.position - v0.position).normalize();
+            let n = n10.cross(n20).normalize();
+            v0.normal = n;
```
philpax, line 53: Could be nice to have an option to calculate the normals automatically from the faces on the host side, but that's out of scope for this PR

philpax (approved):
Awesome work! Glad to see this taking shape 😊

crates/wasm/wit/main.wit
VinhTruongAmbient: Sure. I was thinking name spacing experimental modules under `client-experimental` module, but it `wit` probably doesn't support it?

philpax: I'm not actually sure, give it a shot?

VinhTruongAmbient: I skimmed through https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md, but there doesn't seem to be a way to nest `world` or `interface`, and there are no mentions of modules.

VinhTruongAmbient: Let's go with `client-experimental-X`, where `x=mesh/texture/material`.

FredrikNoren:
Hm why is it in an "experimental" namespace? I'd say everything is experimental at this stage with Ambient :)

VinhTruongAmbient:
> Hm why is it in an "experimental" namespace? I'd say everything is experimental at this stage with Ambient :)

This was @philpax 's idea and I guess the thinking was that this is even more experimental than everything else in Ambient :). I will remove it in the next cleanup pass.

philpax:
Yeah, I was thinking this would be subject to a lot more change than the rest of the API, but I suppose it's reasonable to keep it non-experimental and just make breaking changes as that's what we're doing with the rest of the API.

crates/procedurals/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn new() -> Self {
+        Self {
+            meshes: Default::default(),
+            textures: Default::default(),
+            samplers: Default::default(),
+            materials: Default::default(),
+        }
+    }
+
+    pub fn insert_mesh(&mut self, mesh: ProceduralMesh) -> ProceduralMeshHandle {
```
FredrikNoren, line 107: All of these methods are just redirecting to the underlying values; you can just make them public instead and save on the boilerplate

VinhTruongAmbient: I would like to keep these for now, since we might add more things in the near future, which might require denying free access to the internal state. However if that doesn't happen I might refactor them later.

crates/procedurals/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+    ProceduralTextureHandle,
+};
+use ambient_std::{cb, mesh::Mesh};
+
+components!("procedurals", {
+    @[Resource]
+    procedural_storage: ProceduralStorage,
+});
+
+pub fn systems() -> SystemGroup {
```
philpax, line 31: suggest `client_systems` to make it obvious this only runs on the client for now

crates/procedurals/src/lib.rs
philpax: I'd suggest renaming this to `AllProceduralStorage` (or similar) creating a `ProceduralStorage<ProceduralMaterialHandle, ProceduralMaterial>` struct and making those public, then blocking off access to the internal state, so that you can do `all.textures.get(...)` and ensure they go through the same API without getting direct access and without having to define getters/setters for each resource type.

I think this machinery could also be used for other kinds of procedurals, like #182, so I'd like to reduce the effort required to add a new procedural in future.

crates/wasm/src/client/implementation/specific.rs
```diff
@@ -261,3 +272,147 @@ impl wit::client_window::Host for Bindings {
...
+}
+impl wit::client_material::Host for Bindings {
+    fn create(
+        &mut self,
+        desc: wit::client_material::Descriptor,
+    ) -> anyhow::Result<wit::client_material::Handle> {
+        let world = self.world_mut();
+        let storage = world.resource_mut(procedural_storage());
+        let material = PbrMaterialConfig {
+            source: "Procedural Material".to_string(),
```
philpax, line 389: It might be a good idea to include a `source` in the WIT descriptor, so that users can figure out what material/mesh/etc they're dealing with when looking at a lot of procedural content in a dev tool that shows this

crates/wasm/src/shared/conversion.rs
```diff
@@ -347,3 +348,120 @@ impl FromBindgen for wit::entity::AnimationActionStack {
...
+                }
+            }
+        }
+    };
+}
+
+procedural_storage_handle!(ProceduralMeshHandle, client_mesh);
+procedural_storage_handle!(ProceduralTextureHandle, client_texture);
+procedural_storage_handle!(ProceduralSamplerHandle, client_sampler);
+procedural_storage_handle!(ProceduralMaterialHandle, client_material);
```
philpax, line 393: Might be a good idea to use `paste!` so that you can define the valid handles in one place and share them across modules, a la `primitive_component_definitions`: https://github.com/AmbientRun/Ambient/blob/main/shared_crates/shared_types/src/lib.rs#L3

Not a blocker by any means though, but we may need to document how to add more procedural resources if it's not immediately obvious / there's no one source of truth

crates/wasm/wit/component.wit
```diff
@@ -24,6 +28,10 @@ default interface component {
         type-uvec2(list<uvec2>),
         type-uvec3(list<uvec3>),
         type-uvec4(list<uvec4>),
+        type-procedural-mesh-handle(list<procedural-mesh-handle>),
+        type-procedural-texture-handle(list<procedural-texture-handle>),
+        type-procedural-sampler-handle(list<procedural-sampler-handle>),
+        type-procedural-material-handle(list<procedural-material-handle>),
```
philpax, line 34: I don't love having to extend the list of valid value types for every new procedural handle, but I can't think of a better way to do it at present. We'll likely need to revisit our component value model at some point anyway, so this is fine for now, but it is a bit of a shame

guest/rust/api_core/src/client/material.rs
```diff
@@ -0,0 +1,34 @@
+use crate::global::{ProceduralMaterialHandle, ProceduralSamplerHandle, ProceduralTextureHandle};
+use crate::internal::conversion::*;
+use crate::internal::wit;
+
+#[derive(Clone, Copy)]
+pub struct Descriptor {
+    pub base_color_map: ProceduralTextureHandle,
```
philpax, line 7: I wonder... should these be procedural texture handles, or just texture handles? What if I want to use a texture in my `assets/` with a procedural material?

philpax: Maybe out of scope for this initial version, but I do wonder how we'd handle this

guest/rust/api_core/src/global/procedurals.rs
```diff
@@ -0,0 +1,20 @@
...
+                Self((0, 0))
+            }
+        }
+    };
+}
+
+procedural_storage_handle!(ProceduralMeshHandle);
+procedural_storage_handle!(ProceduralTextureHandle);
+procedural_storage_handle!(ProceduralSamplerHandle);
+procedural_storage_handle!(ProceduralMaterialHandle);
```
philpax, line 20: ditto on finding a way to share these if possible

guest/rust/examples/basics/procedural_generation/src/client.rs
```diff
@@ -0,0 +1,367 @@
+use ambient_api::client::{material, mesh, sampler, texture};
```
philpax, line 1: Collapse all of these into one `ambient_api` import (VSC/R-A has a "merge imports" hint)

guest/rust/examples/basics/procedural_generation/src/client.rs
```diff
@@ -0,0 +1,367 @@
...
+const TAU: f32 = std::f32::consts::TAU;
+const RESOLUTION_X: u32 = 32;
+const RESOLUTION_Y: u32 = 8;
+const TEXTURE_RESOLUTION_X: u32 = 4 * RESOLUTION_X;
+const TEXTURE_RESOLUTION_Y: u32 = 4 * RESOLUTION_Y;
+const SIZE_X: f32 = RESOLUTION_X as f32 / RESOLUTION_Y as f32;
+const SIZE_Y: f32 = 1.0;
+const WAVE_AMPLITUDE: f32 = 0.25;
+const WAVE_FREQUENCY: f32 = 0.5 * TAU;
+const ROTATING_SUN: bool = false;
```
philpax, line 21: Consider making this a runtime option activated on input or with UI

guest/rust/examples/basics/procedural_generation/src/client.rs
```diff
@@ -0,0 +1,367 @@
...
+                    rotation(),
+                    Quat::from_rotation_z(frametime()) * sun_rotation,
+                );
+            }
+        });
+    }
+}
+
+fn make_coordinate_system() {
+    let make_cube = |t: Vec3, s: Vec3, c: Vec4| {
```
philpax, line 59: style nitpick: if these don't close over any style, prefer `fn make_cube(...)` to make that clearer

guest/rust/examples/basics/procedural_generation/src/client.rs
```diff
@@ -0,0 +1,367 @@
...
+        let [i0, i1, i2]: [_; 3] = triangle.try_into().unwrap();
+        let p0 = vertices[i0 as usize].position;
+        let p1 = vertices[i1 as usize].position;
+        let p2 = vertices[i2 as usize].position;
+        let n01 = (p1 - p0).normalize();
+        let n02 = (p2 - p0).normalize();
+        let n = n01.cross(n02).normalize();
+        vertices[i0 as usize].normal += n;
+        vertices[i1 as usize].normal += n;
+        vertices[i2 as usize].normal += n;
```
philpax, line 204: Consider adding an option to `mesh::Descriptor` to have the normals automatically calculated by the host

philpax (approved):
Awesome work! I have a few nitpicks and suggestions for future extensibility/use, but I don't think there are any blockers here. Feel free to resolve whichever ones of these you feel like resolving, we'll make a note of the other ones, and then we'll get this in today 🚀

FredrikNoren (approved):
Took a quick peek and looks good to me, feel free to ping me on discord if there's anything in particular you'd like to discuss though

crates/wasm/src/shared/conversion.rs
VinhTruongAmbient: TIL about `paste`. It did help to cleanup up these macros a lot, thanks!

crates/wasm/wit/component.wit
VinhTruongAmbient: Agreed. I will nuke these once we have a better model in place.

guest/rust/api_core/src/client/material.rs
VinhTruongAmbient: Good question. We briefly discussed this idea where instead of `String` urls, the asset pipeline code generates IDs for each asset, something like `pub const MY_COLOR_TEXTURE: SomeIDType = ...;`. At load time, the engine resolves this ID into the physical location of the asset, which could be in a local directory or at a remote server. This indirection would make asset URLs slightly more efficient since it's no longer an arbitrarily long string, and it would also reduce the chance of human error in typing out string URLs, and also it might be more resilient to directory changes. One downside though is the fact that it is an indirection, so it might be harder to understand.

VinhTruongAmbient: Once we work on custom shaders I think I'll change it to `TextureHandle` to match `BufferHandle`.

guest/rust/api_core/src/client/mesh.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl<'a> IntoBindgen for Descriptor<'a> {
+    type Item = wit::client_mesh::Descriptor<'a>;
+
+    fn into_bindgen(self) -> Self::Item {
+        Self::Item {
+            vertices: unsafe {
+                std::slice::from_raw_parts(
+                    self.vertices.as_ptr().cast::<wit::client_mesh::Vertex>(),
```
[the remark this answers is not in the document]
VinhTruongAmbient: For size I can add a static assert. For type layout I think we can rely on `#[repr(C)]` doing the heavy lifting. Since all participating types are `#[repr(C)]` I think it's ok.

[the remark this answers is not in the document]
VinhTruongAmbient: The best solution would be making `WIT` types implement `bytemuck::Pod` and `bytemuck::Zeroed` if they can.

guest/rust/examples/basics/procedural_generation/src/client.rs
VinhTruongAmbient: Yeah let's do this later. I think we could also add automatically computed UVs (some trivial projection) and tangents (from UVs) too.

PR #397, opened 2023-05-16
Wit binder mask getters

devjobe:
I was planning on making this PR later today but i accidentally submitted it instead of #398, it was ready but i didnt have time to write a proper comment. 

Adds with apis for interacting with animation binder/mask and api for retrieving a set of world transforms.

Adds bone tracking to skinmesh example to show usage.


philpax:
Should I merge both?

devjobe:
Both PR or commits? Yes either way. I made a separate commit of the example changes if its too much for  a basic example or if there are better usage of the Ui  api that you would prefer

philpax:
Hm, maybe move the example to `skinmesh_advanced` - it's definitely something worth having an example for, but it might be overwhelming for people who just want to see the basics.

Otherwise, lgtm, will merge

devjobe:
Moved it to new example skin_tracking and removed the duplicate parts from skinmesh example

PR #409, opened 2023-05-19
fix: Ulid::new() panic on web

ten3roberts:
This resolves the use of Ulid calling `SystemTime::now` which is not implemented on the `wasm32-unknown-unknown` targets, as the `unknown` means that there *is* no libstd or OS, hence time is not available/implemented and just panics.

The way of getting the time is binding to JavaScript using `wasm-bindgen` and calling js `Data.now`.

In addition, the use of `Ulid::now` is now disallowed by clippy in CI

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,14 +1,15 @@
 [target.'cfg(all())']
 rustflags = [
-  # Global lints/warnings.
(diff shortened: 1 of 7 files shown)


VinhTruongAmbient (approved):
Good catch!

Issue #23, opened 2023-02-07
Support project dependencies with relative path


philpax:
Full dependency support is [a lot of work](https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527). For now, we should support relative dependencies and see where that takes us:

```toml
[project]
id = "skinmesh"
name = "Skinmesh"
version = "0.0.1"

[dependencies]
peasant_man = { path = "../peasant_man" }
```

This should make `peasant_man`'s assets and components accessible to `skinmesh`. Additionally, if `peasant_man` introduces any scripts, they should be automatically added to the project, with an opt-out available (this may change if we develop support for spawning scripts from other scripts).

PR #411, opened 2023-05-20
Primitives : Simple Procedural trees ( pseudo randomness included ) 

mebyz:
This PR adds a TreeMeshClass that represents a simple procedural tree with trunk and foliage. it also adds an advanced example to show the usage and build up a golden test.

Capture d’écran 2023-05-21 234720

- implements PSRNG (pseudo random number generator) :  use a seed to create unique trees that will be rendered the same for all platforms, tests, clients, etc.

- build up entire forests from a single seed :)

thanks a lot for your kind reviews 👍


FredrikNoren:
@mebyz Hey this is super cool! Great work

However, with #391 and #23, I think it would make more sense to have this as a "mod"/"plugin"/"dependency" (name pending hehe), rather than keeping it in the engine itself. We're also planning to move out other things from the engine into mods as soon as #23 lands. For now I'd recommend setting it up as a new github repository, and then we can work on making it possible to depend on it from other projects.

crates/std/src/uncategorized/asset_url.rs
```diff
@@ -436,11 +436,20 @@ impl AssetUrl {
             AssetUrl::Absolute(url) => Ok(Self::Absolute(url.join(path)?)),
             AssetUrl::Relative(p) => Ok(Self::Relative(
                 // The Url::join method has some intricacies so we want these to behave the same
-                Url::parse(&format!("http://localhost/{}", p.as_str()))
+                if cfg!(target_os = "windows") {
```
mebyz, line 439: sadly on windows, when using a pipeline.json file that contains more than one pipeline like this : 

`
[
    {
        "pipeline": {
            "type": "Materials",
            "importer": {
            "type": "Single",
            "name": "sand",
            "base_color": "sand.png"
            },
            "output_quads": true
        }
    },
    {
        "pipeline": {
          "type": "Materials",
          "importer": {
            "type": "Single",
            "name": "bark",
            "base_color": "bark.png"
          },
          "output_quads": true
        }
      }
]
`

app fails at runtime because it can't save the material asset(s) file ie : c:\\ambient\\guest\\rust\\examples\\advanced\\forest\\build\\assets\\pipeline.json/0/mat.json

because path mixes slashes and backslashes

i've sumbitted a PR on udoprog/relative-path here https://github.com/udoprog/relative-path/pull/43/files

so we don't maintain this platform-specific code

mebyz:
@FredrikNoren thanks for your message, from what i understand, what i need to do now : 

- create a new repo (or maybe you create that repo in ambientrun org ?)
- move the TreeMesh class code to that repo 
- make a crate of that code so it can be used as a dependency
- try to use that crate as a dependency in my advanced/forest example, (which would remain in ambient repo ?)

i'm unsure at that point on what to do with the specific components infos that TreeMesh uses (such as foliage_radius, or the tree component itself). for now they are declared in the .\shared_crates\schema\src\schema\primitives.toml file. where should these go ?

did i miss anything? thanks a lot :)

FredrikNoren:
@mebyz Yeah exactly. The example would also live in the new repository. It wouldn't be a crate dependency, but an "ambient dependency" though (which doesn't exist yet, but it's under way).

For the component infos; you would just use an ambient.toml in that project to define them.

Like I said, everything isn't implemented yet so it will take a little bit of time before people can use your code as a dependency, but hopefully soon!

mebyz:
Thanks again @FredrikNoren, if i Can be of any help on the ambient deps subject, just let me know :) hopefully this use case can serve as an early test for the New dependencies system you mention.

FredrikNoren:
@mebyz Alright, I'll close this PR now since we have a path forward without it

PR #414, opened 2023-05-22
Add conditionally bundled certificates and cli certificate overrides

ten3roberts:
This adds a bundled (include_bytes) certificate and key that can be disabled in production using `no_bundled_certs` (default with `production` feature.

This also allows ambient to be used with our self-signed certificates outside the ambient repo root as it would previously fail to find the cert file and key.

See the modified docs for usage 😊 

If a cert file,key, or ca is specified it will not fall back on the bundled certificates if they fail to load to avoid a false trust security vulnerability if the file is silently removed.

### DIFF
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -43,7 +43,9 @@ ambient_ecs_editor = { path = "../crates/ecs_editor" }
 ambient_editor_derive = { path = "../shared_crates/editor_derive" }
 ambient_element = { path = "../shared_crates/element" }
 ambient_project = { path = "../shared_crates/project" }
-ambient_shared_types = { path = "../shared_crates/shared_types", features = ["native"] }
+ambient_shared_types = { path = "../shared_crates/shared_types", features = [
+    "native",
+] }
 
 tracing-tree = { workspace = true, optional = true }
 tracing-subscriber = { workspace = true, optional = true }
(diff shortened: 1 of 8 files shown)


app/Cargo.toml
```diff
@@ -76,9 +78,10 @@ glam = { workspace = true }
 rusty-hook = "^0.11.2"
 
 [features]
-default = []
+no_bundled_certs = []
+default = ["assimp"]
```
pierd, line 82: Did this sneak in via merge?

ten3roberts: I conflicted with another commit, will go over this

ten3roberts: Yes, good catch, it was a result of the merge with Fredrik's commit to main and git auto merge.

Is fixed now. This was moved from `default` to `production` on main, but auto merge added both

pierd (approved):

FredrikNoren:
@ten3roberts @pierd Hm doesn't the release version of ambient need the bundled certs to work as well? If I download ambient.exe from github releases I mean.

ten3roberts:
This was something I was thinking about last night.

Namely how do we tell apart someone who wants to use the downloaded client to connect to our servers and play the secure way, versus someone who uses the downloaded client to host their own game and join it locally.

The certs are not bundled as in production environments you should not use self-signed certs as it defeats the purpose of TLS and makes man in the middle attacks very possible (since the key of the cert is publicly available in our repo).

I think the question boils down to:

*How do we tell apart a client that is used for production and deployed purposes, versus a client which is used for development purposes*

FredrikNoren:
@ten3roberts Hm can't you just assume it's for development, and for production we can have a command line arg to override the certificate?

ten3roberts:
I am not exactly sure what you mean by *override certificate*?

PR #417, opened 2023-05-23
Semantically-backed project manifests with dependencies

### LINKED ISSUE #23: Support project dependencies with relative path

### LINKED ISSUE #107: Allow the use of `core` components and concepts in user-defined concepts
At present, concepts are restricted to using components / extending concepts defined in the same `ambient.toml`. The API macro crate includes the core `ambient.toml`, so it should be able to draw upon it to find components and concepts to use if not defined in the current project.

This needs to:
- resolve the paths to the components and concepts from the `core` module correctly
- work with existing use-cases
- not import those concepts into scope
- be implemented in such a way that it could be extended to support using components/concepts from project dependencies (#23) in future.

### LINKED ISSUE #221: Enum's for `ambient.toml` components
The `ambient.toml` components currently only support primitive types, but a lot of APIs require enum type data. Perhaps something like this:

```rust
[project]
id = "my_project"
name = "My project"
version = "0.0.1"

[types]
PlayerState = { enum = ["Alive", "Dead", "Dying"] }

[components]
player_state = { type = "PlayerState", name = "Player state" }
```

### LINKED ISSUE #405: Preserve order of fields for things defined in the manifest
The manifest is loaded through the `toml` and `serde` crates, with the use of `BTreeMap` to store data. However, this will destroy the order of items as they were specified in the manifest. This is especially annoying with messages, whose fields will have a different order in the `new` constructor.

It looks like there's a way to use a `LinkedHashMap` when deserializing to preserve order: https://docs.rs/toml/latest/toml/type.Table.html

but I'm not sure if this is doable while still deserializing the entire structure with `serde`. I will have to look into this.

### LINKED ISSUE #533: UI align end in examples doesn't work
image

### LINKED ISSUE #538: Descriptive error message for invalid types in components and message fields
```
[messages.player_command.fields]
mvx = { type = "I8" }
mvy = { type = "I8" }
```

The error message that results from this message field definition (in ambient.toml) is currently the following:

```
called `Result::unwrap()` on an `Err` value: invalid primitive type
```

Instead it should say something like this:

```
Message field type definition in ambient.toml (line 7) is invalid (type "I8" is unsupported by Ambient protocol)
```

As many of the below as possible, to assist the user in understanding the user error aspect of the issue, so that they can resolve it quickly by themselves

- Refer to problematic filename (e.g. 'ambient.toml')
- Refer to line number in file (not that important, but nice if it is possible)
- Describe the problem at the user error level (e.g. 'message field type definition is invalid') rather than in terms of the backend problem (i.e. I have not written any `Result::unwrap` code so I have no idea what it is referring to, or how to fix it)
- Use language that indicates that the issue is not a bug, but known and intentional ('unsupported by Ambient')

philpax:
This adds a semantic layer to embers, allowing them to reference each other, provide much better diagnostics, and more. This is a "compiler" for ember semantics that is integrated in both the runtime and the ember macro. Deployments and remote dependencies will require "package management".

Adds:
- enums
- dependencies on other projects
- being able to use `core` components in your own concepts etc
- the determination of issues with missing items at the manifest level, not at the Rust level

Closes #221, closes #107, closes #405, closes #23, closes #533, closes #538.

**TODO**:
- [x] do an actual topo-sort of scopes
- [x] error on dependency cycles
- [x] reserve keywords that can't be used as identifiers ("client", "server", "ember", "component", "message", "concept", "enum", etc)
- [x] make enums work for third-party components
- [x] make deployment for single-ember projects work again
- [x] make loading a single-ember game/ember from a URL work again
- [x] check that the right ID is being used for scopes
- [x] think about whether the root asset should go in build/ or build/its_name
  - [x] conversely, figure out what to do about the existing examples - maybe `asset::url` and then each dependency gets `dep::asset::url`?

**TODO to convert to issues**:
- [x] update changelog
- [x] update documentation
- [ ] make deployment for multi-ember projects work
- [ ] make loading a multi-ember game from a URL work
- [ ] make runtime ember loading possible
- [ ] make runtime ember disabling possible
- [ ] switch to friendly IDs for ember names, use names across the board
- [ ] when dealing with a hierarchy that has multiple renamed names for the same scope, merge the scopes together in the macro and use aliases
- [ ] rename project to ember across the board
- [ ] move messages to more appropriate locations
- [ ] add concepts for core components (character controller, etc)
- [x] only build changed embers


philpax:
rip, broke the PR with filter-repo shenanigans, will try to fix

philpax:
Known issues:
- enums don't work when defined by users.
  - handle creating the enum attribute for ExternalComponentAttributes. Might need to revisit how ExternalComponents work now that we have "virtual types".
- dynamic ember loading isn't supported yet, but when it is, it should probably have some kind of progress indicator

shared_crates/schema/build.rs
```diff
@@ -13,9 +13,10 @@ fn main() {
             (
                 path.strip_prefix("src")
                     .unwrap()
-                    .to_str()
-                    .unwrap()
-                    .to_owned(),
+                    .iter()
```
ten3roberts, line 16: I do recall us explicitely *not* doing a lossy string conversion to avoid having multiple dirs mapping to the same value if not UTF-8 (in an earlier PR)

philpax: This is only used to create a lookup table for the schema which we control, so I'm not too concerned.

PR #420, opened 2023-05-24
Golden image test improvements

VinhTruongAmbient:
## Golden image tests are now migrated into Campfire utility

Users no longer need to install `node` just to run golden image tests. New command looks like: `cargo cf golden-images <command>`.

## Improved reliability

Previously, golden image tests used `std::process::exit()` to exit the program after the test is done. This way of exiting the process caused seemingly random golden image failures, even when both before and after images are identical. Just the act of calling `exit()` seems to be undefined behavior in multithreading environments.

New implementation simply exits gracefully by bubbling some state up the callstack. This change should improve test reliability.

## Separated `update` and `check` commands

Previous implementation implicitly updated the golden image if the comparison failed for the `timeout`-number of seconds. New implementation allows you to explicitly update all images, effectively overriding any previous images, and separately check against existing images. Check is allowed to fail if the existing image does not exist. 

New commands:
```
cargo cf golden-images update
cargo cf golden-images check
```

`update` is specified with `--wait-seconds` which will wait for the specified number of seconds before capturing and saving the current image, and then exiting the program.

## Progress bars

Running this tool in an interactive terminal (non-CI environments), there's now a progress bar that approximates how long the tests take to finish.

Screenshot 2023-05-25 at 14 22 50


philpax (approved):
heck yeah, glad it was pretty straightforward to do

.github/workflows/ci.yml
```diff
@@ -110,32 +110,26 @@ jobs:
...
       - name: Install run (headless) dependencies
         run: |
           sudo add-apt-repository ppa:oibaf/graphics-drivers -y
           sudo apt-get update
           sudo apt install -y libxcb-xfixes0-dev mesa-vulkan-drivers
 
-      - name: Cargo cache
-        uses: actions/cache@v3
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
```
FredrikNoren, line 124: Won't this use the same cache for this build as the other builds?

VinhTruongAmbient: With this setup only golden image runs share caches with each other. But for example test-linux job does not. Here I am experimenting if this cache implementation does a better job than what we had before, so I didn't change other jobs yet.

philpax: Did we decide if this was better or not?

.github/workflows/ci.yml
```diff
@@ -110,37 +110,35 @@ jobs:
           sudo apt-get update
           sudo apt-get install --no-install-recommends -y tree libasound2-dev libglib2.0-dev libxcb-shape0-dev libxcb-xfixes0-dev \
             libcairo-dev libgtk2.0-dev libsoup2.4-dev libgtk-3-dev libwebkit2gtk-4.0-dev xorg-dev ninja-build libxcb-render0-dev
+      - name: Install protoc
```
philpax, line 113: Is this still necessary since this doesn't do a workspace build?

app/src/cli/mod.rs
```diff
@@ -70,6 +70,20 @@ pub enum Commands {
     },
 }
 
+#[derive(Subcommand, Clone, Copy, Debug)]
+pub enum GoldenImageCommand {
+    /// Renders an image and saves it after waiting for the specified number of seconds
+    GoldenImageUpdate {
```
philpax, line 76: extremely minor nit: can drop the `GoldenImage` in the name of these variants, it's implied by being part of `GoldenImageCommand`

campfire/src/golden_images/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+use std::{process::Command, time::Instant};
+
+mod failure;
+use failure::*;
+
+mod progress;
+use progress::*;
+
+const TEST_BASE_PATH: &str = "guest/rust/examples";
+const TESTS: &[&str] = &[
```
philpax, line 17: Is there a way to avoid hardcoding this? Maybe have a file at the root of `guest/rust/examples` that lists the projects to test?

philpax (approved):
lgtm, small nitpicks but everything else looks sound to me

.github/workflows/ci.yml
VinhTruongAmbient: Sorry, forgot to update this thread. Yeah I think it works quite well and it is less definitions than the previous one. It was recommended by the author of rust-analyzer and they use it too.

.github/workflows/ci.yml
VinhTruongAmbient: I don't think it is, will remove.

campfire/src/golden_images/mod.rs
VinhTruongAmbient: I can add a manifest under `guest/rust/examples` which lists all tests.

campfire/src/golden_images/mod.rs
```diff
@@ -14,29 +15,16 @@ mod progress;
...
-    "ui/flow_layout",
-    "ui/rect",
-    "ui/screens",
-    "ui/slider",
-    "ui/text",
-];
+const TEST_MANIFEST: &str = "golden-image-manifest.toml";
+
+#[test]
+fn lol() {
```
philpax, line 21: Still necessary?

VinhTruongAmbient: Derp, fixed.

PR #423, opened 2023-05-24
v0.3.0-dev

philpax:
Updates all versions and fixes a bug in Campfire where partial version updates would result in the dependency graph not being built correctly.

Will merge once CI passes.


ten3roberts:
Great, will update my webtransport PR soon to accomodate this

Issue #424, opened 2023-05-24
Text rendering of zero sized meshes broken


FredrikNoren:
@VinhTruongAmbient

philpax:
Is this related to #183 ?

VinhTruongAmbient:
This issue is related to https://github.com/AmbientRun/Ambient/pull/393 where I validated meshes. However this PR introduced a subtle bug here:
https://github.com/AmbientRun/Ambient/blob/main/crates/text/src/lib.rs#L412-L415
It is possible for text code to generate a mesh with no vertices. I don't know yet why this happens, but I feel like this should not happen in the first place.

FredrikNoren:
@VinhTruongAmbient Think it does that when there are 0 characters in the text

PR #425, opened 2023-05-24
Fix issue #424

VinhTruongAmbient:


philpax (changes requested):
`remove_component` and `remove_components` are idempotent so you should be able to do `world.remove_components(id, vec![primitives().desc(), mesh().desc()])` instead of checking for their presence

VinhTruongAmbient:
> `remove_component` and `remove_components` are idempotent so you should be able to do `world.remove_components(id, vec![primitives().desc(), mesh().desc()])` instead of checking for their presence

Gotcha. What about the returned `ECSError`? If a component is missing, it will return `ECSError::EntityDoesntHaveComponent` right?

philpax:
Nope, I don't think that error is used for removing. It's a bit confusing, I admit.

FredrikNoren (approved):

PR #427, opened 2023-05-24
Fix TLS handling for proxy connection

pierd:


crates/network/src/native/mod.rs
```diff
@@ -3,3 +3,44 @@
 //! This included quinn server+client and webtransport server using `h3`
 pub mod client;
 pub mod server;
+
+#[tracing::instrument(level = "info")]
+fn load_native_roots() -> rustls::RootCertStore {
```
ten3roberts, line 8: Perhaps it would be easier make two separate functions, one for each feature, to alleaviate the interleaving of cfg guards.

Additionally, if both features are active, which can easily happen due to cargo feature unification, you create two roots stores and shadow the first one.

How does it compile without mut in the `tls-webpki-roots` feature?

ten3roberts (changes requested):

crates/network/src/native/mod.rs
ten3roberts: I was wrong, that non mut root is when *neither* feature is active

pierd: I'm worried that splitting into separate functions just pushes the decision up. Unless you mean adding separate function for each store?

Using the sum of 2 stores should be fine (as in correct - security-wise less so).

The `mut` is clumsy here because `roots` have to be mutable if any store is used but if there isn't any then clippy doesn't like it.

ten3roberts: I've usually found that splitting into separate functions for each cfg has worked that best, rather than having the cfg flags inside the functions.

That is the pattern I've used with great success in *ambient_sys*

pierd: I was thinking about something like this:
```rs
fn add_native_roots(roots: &mut rustls::RootCertStore) {
    #[cfg(feature = "tls-native-roots")]
    {
        match rustls_native_certs::load_native_certs() {
            Ok(certs) => {
                for cert in certs {
                    let cert = rustls::Certificate(cert.0);
                    if let Err(e) = roots.add(&cert) {
                        tracing::error!(?cert, "Failed to parse trust anchor: {}", e);
                    }
                }
            }

            Err(e) => {
                tracing::error!("Failed load any default trust roots: {}", e);
            }
        };
    }
}

fn add_webpki_roots(roots: &mut rustls::RootCertStore) {
    #[cfg(feature = "tls-webpki-roots")]
    {
        roots.add_server_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0.iter().map(|ta| {
            rustls::OwnedTrustAnchor::from_subject_spki_name_constraints(
                ta.subject,
                ta.spki,
                ta.name_constraints,
            )
        }));
    }
}

#[tracing::instrument(level = "info")]
fn load_root_certs() -> rustls::RootCertStore {
    tracing::info!("Loading root certificates");

    let mut roots = rustls::RootCertStore::empty();
    add_native_roots(&mut roots);
    add_webpki_roots(&mut roots);
    roots
}
```

So still inside the functions. It still doesn't make clippy happy because `roots` argument is not used if given features is not enabled.

pierd: Or actually a bit cleaner:

```rs
#[cfg(feature = "tls-native-roots")]
fn add_native_roots(roots: &mut rustls::RootCertStore) {
    tracing::info!("Loading native root certificates");
    match rustls_native_certs::load_native_certs() {
        Ok(certs) => {
            for cert in certs {
                let cert = rustls::Certificate(cert.0);
                if let Err(e) = roots.add(&cert) {
                    tracing::error!(?cert, "Failed to parse trust anchor: {}", e);
                }
            }
        }

        Err(e) => {
            tracing::error!("Failed load any default trust roots: {}", e);
        }
    };
}

#[cfg(feature = "tls-webpki-roots")]
fn add_webpki_roots(roots: &mut rustls::RootCertStore) {
    tracing::info!("Loading webpki root certificates");
    roots.add_server_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0.iter().map(|ta| {
        rustls::OwnedTrustAnchor::from_subject_spki_name_constraints(
            ta.subject,
            ta.spki,
            ta.name_constraints,
        )
    }));
}

#[tracing::instrument(level = "info")]
fn load_root_certs() -> rustls::RootCertStore {
    #[cfg(any(feature = "tls-native-roots", feature = "tls-webpki-roots"))]
    {
        let mut roots = rustls::RootCertStore::empty();
        #[cfg(feature = "tls-native-roots")]
        add_native_roots(&mut roots);
        #[cfg(feature = "tls-webpki-roots")]
        add_webpki_roots(&mut roots);
        roots
    }
    #[cfg(not(any(feature = "tls-native-roots", feature = "tls-webpki-roots")))]
    {
        tracing::info!("Creating empty root certificates store");
        rustls::RootCertStore::empty()
    }
}
```

ten3roberts: I do the second one a *lot*.

What I also tend to do is use a very local `#[allow(unused_variables)]` instead of the more brittle any of feature combination, which allows me to only use `cfg(feature)` without any `any` or `all` or other boolean operators in the cfg directive.

Unnused in rust is mostly fine, as Vec::new and such don't do anything, so the warning is purely to remind *you* to not forget anything. 

I usually do that trick when writing tuple impl macros

ten3roberts: Another trick instead of an `allow` directive, is to just name it `_roots`, which makes cargo happy *and* still allows you to use the variable

PR #433, opened 2023-05-27
Perf: Low hanging fruit

devjobe:
Changes:
* Removed old animation blend query
* Removed duplicate work in fbx_complex_transform query
* Refactored fbx_complex_transform

Test: 400 instances of single animated character with fbx_complex_transforms:

FPS: 20 -> 40
Overall improvement: 50~ ms / frame -> 24~ ms / frame
animation_systems 15.5 ms -> 7.5 ms
TransformSystem::run 29 ms -> > 12.5 ms

Before:
[Before](https://github.com/AmbientRun/Ambient/assets/71590722/dcd540d2-d02f-4f6c-a589-d42285a6dfe1)

After:
[After](https://github.com/AmbientRun/Ambient/assets/71590722/250c3d44-f873-4d78-8ebe-f0439b6a0d4b)

fbx_complex_transform is now about 3ms of the work, the rest is transform propagation. I would probably remove completely it unless someone has good reason for it to stay. 

Perf is not horrible but the animation sampling as a lot of room for improvement. The transform system I'm not so sure about but obviously all of this could be parallelized.

### DIFF
--- a/crates/animation/src/lib.rs
+++ b/crates/animation/src/lib.rs
@@ -1,6 +1,6 @@
 use ambient_core::{asset_cache, hierarchy::children, time};
 use ambient_ecs::{
-    components, query, Debuggable, EntityId, MakeDefault, Networked, Store, SystemGroup,
+    components, query, Debuggable, EntityId, MakeDefault, Networked, Store, SystemGroup, World,
 };
 use ambient_model::{animation_binder, model, model_from_url, ModelFromUrl};
 use ambient_std::{
@@ -137,10 +137,6 @@ fn anim_time(
 }
 
 impl AnimationAction {
-    fn time(&self, time: &Duration, clip: &AnimationClip) -> f32 {
-        anim_time(time, &self.time, clip, self.looping)
-    }
-
     fn sample_tracks(
         &self,
         time: &Duration,
@@ -210,13 +206,6 @@ impl AnimationController {
     }
 }
 
-#[derive(Debug)]
-struct AnimationBlendOutput {
-    target: EntityId,
-    value: AnimationOutput,
-    weight: f32,
-}
-
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum AnimationActionStack {
     Sample {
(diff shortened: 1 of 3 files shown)


philpax (approved):
Animation stuff looks good, not too sure about the intricacies of the FBX stuff. That'd be up to @FredrikNoren to review.

crates/core/src/transform.rs
```diff
@@ -277,48 +277,122 @@ impl TransformSystem {
...
-                        .to_system(|q, world, qs, _| {
-                            // See: https://help.autodesk.com/view/FBX/2017/ENU/?guid=__files_GUID_10CDD63C_79C1_4F2D_BB28_AD2BE65A02ED_htm
-                            // and: https://github.com/assimp/assimp/blob/add7f1355e96c6ff0df0ba3cec084f25332d154e/code/AssetLib/FBX/FBXConverter.cpp#L687
-                            for (id, _) in q.collect_cloned(world, qs) {
-                                world
-                                    .set(id, local_to_world(), get_fbx_transform(world, id))
-                                    .unwrap();
-                            }
-                        }),
+                    query_mut(
```
FredrikNoren, line 280: I'm not sure that they will always have _all_ of these components, which was why it was implemented as it was.

FredrikNoren: Oh I see, you're always setting them now. Yeah I guess that's fine

crates/animation/src/lib.rs
```diff
@@ -239,6 +228,61 @@ pub enum AnimationActionStack {
...
+            .unwrap();
+    }
+
+    if !world.has_component(id, animation_binder_weights()) {
+        world
+            .add_component(id, animation_binder_weights(), vec![])
+            .unwrap();
+    }
+
+    let mut stack = vec![];
```
FredrikNoren, line 244: What's the relationship between AnimationAction and the AnimationStack?

crates/core/src/transform.rs
devjobe: Yeah, I did an unsound optional variant that got a modest ~25% increase in performance. I think in the case of transforms it should be worth to spend a little bit more on memory to utilize the hardware better.

crates/animation/src/lib.rs
devjobe: AnimationStack operates on AnimtionActions by index, the idea is that you could preload the animations and then blend them as you need every frame. I'm hoping this will also make it straightforward to extend the blending with procedural poses that produces a consistent result if nothing else.

I'm not sure this is the end game of the animation system but what I'm trying to be considerate about at this moment is the amount of data that needs to be transferred from guest every frame. As that's what I'm hearing is problematic with WASM in the browser. It may be that guest won't even interact with this system at this "low level" and then it might not even matter.

FredrikNoren: Hm isn't AnimationActions and AnimationStack doing the same thing then? AnimationActions are a list of animations with parameters of how they should be blended together; AnimationStack seems to be doing the same thing? Or is there a reason why both are needed? (It's quite a while since I looked at this though)

For pre-loading we should just be able to `AnimationClipRef::XX {].peek()` to preload that animation.

FredrikNoren:
For the performance tests; were they run in release or debug?

devjobe:
> For the performance tests; were they run in release or debug?

I used `cargo build --release --features "profile" ` and puffin_viewer

crates/animation/src/lib.rs
devjobe: They are, the difference being one is doing stack operations with option doing a masked blend. AnimationStack makes it straightforward to produce the right order of operations. 

I tried to use AnimationActions todo the same things a few weeks ago but it felt confusing to get the right blend when transitioning.

As for the preloading, I'm thinking more like a AnimationSet rather than individual clips.

ten3roberts:
Now that the components are all added, is there any increased memory consumption, I see most are vecs and quats, so it is at least 144 more bytes, which for 10 fbx objects is roughly an additional 1kb.

If my undestanding of the current animation system is correct this also applies to every bone on a rigged mesh, so a single model with limbs can be 1kb more in memory.

Though I may be incorrect, please do let me know

devjobe:
> Now that the components are all added, is there any increased memory consumption, I see most are vecs and quats, so it is at least 144 more bytes, which for 10 fbx objects is roughly an additional 1kb.
> 
> If my undestanding of the current animation system is correct this also applies to every bone on a rigged mesh, so a single model with limbs can be 1kb more in memory.
> 
> Though I may be incorrect, please do let me know

For some reason it's using less memory after the PR which I don't immediately understand why.  I believe my test is 20 000 objects but I'd have to check, which would mean 2-3 MB of almost unnecessary memory, continuing your math it would be ~144 mb for a million objects. 

I don't know what the ramifications of adding additional entities instead which I believe @FredrikNoren said was the alternative. But I would remove fbx_complex transform if not for performance reasons which doesn't seem as bad as I first thought but from a user perspective there are numerous ways to effect the local_to_parent transform.

I can only imagine someone debugging issues to find out the reason their code wasn't working is because they were using procedural meshes before and fbx after.

I would have to look at it closer, but I would probably go the traditional route and replace local_to_parent with Affine3A or something similar and remove all other components and replace them with utility methods.

FredrikNoren:
@devjobe Re. only having matrix components: just having one true place where the data lives would be amazing, but the challenge would be all the systems that need to work with subsets of that data (ie animation, layout etc which might “just” work with translation for instance). Probably doable though, but a bit of a fun challenge ;)

FredrikNoren:
@devjobe I’ll merge this now. If you have some time to clean up the animation action/ stack stuff I’d also appreciate it (so that it’s in a singular, consistent state)

Issue #432, opened 2023-05-26
Configurable texture filtering in materials asset pipeline


marceline-cramer:
I have some pixel art textures I'm loading using the materials pipeline (a la the `image` example) but they're drawn with linear filtering by default. Can we add the option for nearest-neighbor filtering in the material pipeline?

philpax:
One for you to tackle if you have some free time, @VinhTruongAmbient ?

I think you'd have to extend this with an option for a sampler:

https://github.com/AmbientRun/Ambient/blob/ae54f3e0fbff4b6a7b7c093ebd10441d6cb226a2/crates/build/src/pipelines/materials/mod.rs#L115-L152

and then wire it up here:

https://github.com/AmbientRun/Ambient/blob/ae54f3e0fbff4b6a7b7c093ebd10441d6cb226a2/crates/renderer/src/materials/pbr_material.rs#L443

@marceline-cramer would a simple `Linear`/`NearestNeighbor` sampler choice be sufficient, or would you want full control over the sampler? The latter is better in the long run but it might require more work with asset keys/the cache.

VinhTruongAmbient:
This should be a simple thing to add. While we are at it we should also add wrap modes to be complete.

marceline-cramer:
> @marceline-cramer would a simple `Linear`/`NearestNeighbor` sampler choice be sufficient, or would you want full control over the sampler? The latter is better in the long run but it might require more work with asset keys/the cache.

It would be sufficient, but there's no rush to get this in. I'll need it eventually but in my opinion it's better to prioritize the long run.

VinhTruongAmbient:
My #440 solved this in the most simple way possible. I created a follow-up issue #443 that tackles arbitrary texture-sampler pairs, which is going to be a more involved project.

PR #440, opened 2023-05-30
Overridable sampler in asset pipeline

VinhTruongAmbient:
A simple implementation of #432. Limitations of this design:

- The sampler mode can only be specified with the `material_overrides` system. Any sampler definitions in GLTF or FBX are ignored currently because of the next limitation.

- You can not specify sampler modes per texture. This requires a more significant change in the renderer which we will defer to a later time.


FredrikNoren (approved):
Nice!

PR #444, opened 2023-05-30
Fallible asset pipeline with more precise errors

ten3roberts:
```
[2023-05-30T12:26:05Z INFO  ambient] Building Asset Loading
Error: Failed to build project

Caused by:
    0: Failed to proccess pipelines for "/Users/teiroberts/dev/Ambient/guest/rust/examples/basics/invalid_asset_loading/assets"
    1: Failed to read pipeline "/Users/teiroberts/dev/Ambient/guest/rust/examples/basics/invalid_asset_loading/assets/pipeline.json"
    2: data did not match any variant of untagged enum PipelineOneOrMany

Stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/teiroberts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/libunwind.rs:93:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/teiroberts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/mod.rs:66:5
   1: backtrace::backtrace::trace
             at /Users/teiroberts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.67/src/backtrace/mod.rs:53:14
   2: anyhow::backtrace::capture::Backtrace::create
             at /Users/teiroberts/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.70/src/backtrace.rs:216:13
   3: anyhow::backtrace::capture::Backtrace::capture
             at /Users/teiroberts/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.70/src/backtrace.rs:204:17
```


ten3roberts:
Absolutely

ten3roberts:
An additional change in this PR is a top-level block on (using `#[tokio::main]`) rather than spurious short lived `block_on`s.

This makes the code more clear as awaits can be done in main, as well as task spawning.

app/src/cli/mod.rs
```diff
@@ -70,6 +70,19 @@ pub enum Commands {
         /// The server to connect to; defaults to localhost
         host: Option<String>,
     },
+    /// Asset manipulation and migration
```
FredrikNoren, line 73: Hm we should strive for making the ambient code base as small as possible. Can you put the migrate stuff in a separate repository? Or just remove it, I don't think anyone has a complex pipeline yet so people can just migrate by hand

FredrikNoren (changes requested):
What about the reference documentation? That's all for json: https://ambientrun.github.io/Ambient/reference/asset_pipeline.html#reference
We need to migrate that to toml as well

ten3roberts:
> What about the reference documentation? That's all for json: https://ambientrun.github.io/Ambient/reference/asset_pipeline.html#reference We need to migrate that to toml as well

Entirely forgot about that part, thank you.

app/src/cli/mod.rs
ten3roberts: Mithun mentioned that it'd be best to have it in ambient since others may need to use it when developing.

So his conclusion was to have it as part of our tool

philpax: I think we should keep it in here for now and then delete it for the next release

FredrikNoren: They can use it even if it's in a separate repo. We need to make sure the codebase doesn't balloon in size with temporary code patches like this. Each line of code is adding build times, it's more code to maintain, it makes it harder to understand (migrate will only be relevant for a few weeks now, then never again).

As a general principle, we should try to minimize the number of lines of code in the codebase. Each line of code should be seen as a cost, so it needs to have a lot of long term benefit to warrant a place in the code base; otherwise it shouldn't be written or it should live in an outside repo.

cc @philpax

FredrikNoren: > I think we should keep it in here for now and then delete it for the next release

Sure that's fine, but @ten3roberts you're on the hook for making sure that happens. Set a calendar reminder for yourself to remove it

philpax:
> > > What about the reference documentation? That's all for json: https://ambientrun.github.io/Ambient/reference/asset_pipeline.html#reference We need to migrate that to toml as well
> > 
> > 
> > That's auto-generated in campfire, you can try building a toml outputter instead
> 
> Was not aware of that. How would we go about keeping this reference up to date with Rust and vice-versa, now we just have to independent references that will keep falling out of sync. I can look into having serde output a reference.
> 
> Additionally, toml does not support everything that json does so they can not (always) be freely converted.
> 
> I mentioned it on discord as well, but the current docs on main regarding the pipelines examples in the markdown are not valid.

On my phone but check out

https://github.com/AmbientRun/Ambient/blob/main/campfire/src/doc/mod.rs#L67
https://github.com/AmbientRun/Ambient/blob/main/campfire/src/doc/typescript.rs

The examples will need to be hand done but you should be able to write a toml.rs that takes the AST and generates valid TOML.

ten3roberts:
The reference is now kept up to date by a combined use of rustdoc and testing of the asset pipeline in our examples. The example pipeline code is included in the marrkdown by mdbook directly, so that it never goes stale as long as golden images pass.

FredrikNoren (approved):

PR #450, opened 2023-05-30
improve spatial audio api on physics example

chaosprint:
- [ ] figure out the best way to have the spatial audio API
- [ ] add a spatial audio example with a movable object that plays sound
- [ ] docs


FredrikNoren:
@chaosprint Can we create a separate example for audio/spatial audio? I want to try to keep the examples as small as possible, and for a singular concept.

PR #453, opened 2023-05-31
`Gpu` ownership cleanup in `gpu` and `rendering` crates

ghost:
The primary goal of this PR is to improve the readability of functions which interact with the graphics API. Previously, most GPU-related types cloned the `Arc<Gpu>` and stored locally next to their actual data. Then the member functions were able to access the `Gpu` via `self.gpu`. While this can save some typing, it obscures the fact that a function might be doing something noteworthy with the graphics API, like writing data into a GPU buffer.

We can achieve this goal by changing the "shared ownership of `Arc<Gpu>`" into just a simple `&Gpu`. In practice this means removing almost all `Arc<Gpu>` from structures, and instead passing `&Gpu` as a function argument down the callstack. In most cases the `Gpu` already exists somewhere higher in the callstack, so it doesn't need to be fetched again. A notable exception is all the AssetKey-types where you must get the `Arc<Gpu>` from the asset cache.

The only three places where `Arc<Gpu>` still lives are:

- `AppResources`
- `World`
- `AssetCache`

This PR shouldn't cause any functional changes to the renderer.

### DIFF
--- a/app/src/client/mod.rs
+++ b/app/src/client/mod.rs
@@ -3,7 +3,7 @@ use std::{collections::HashMap, path::PathBuf, sync::Arc, time::Duration};
 use ambient_app::{fps_stats, window_title, AppBuilder};
 use ambient_cameras::UICamera;
 use ambient_core::{
-    runtime,
+    gpu, runtime,
     window::{
         cursor_position, window_ctl, window_logical_size, window_physical_size,
         window_scale_factor, ExitStatus, WindowCtl,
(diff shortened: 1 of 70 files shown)


FredrikNoren (approved):

FredrikNoren:
lgtm. Think this was how we had it in the beginning but we got tired of passing gpu everywhere :) But doesn't really matter, either way is fine for me

VinhTruongAmbient:
> lgtm. Think this was how we had it in the beginning but we got tired of passing gpu everywhere :) But doesn't really matter, either way is fine for me

Thanks. And this does matter a lot to me at least as someone who is not as familiar with the renderer code 🙂.

PR #455, opened 2023-05-31
Add caching to `clippy` and `clippy-web` CI jobs

VinhTruongAmbient:


ten3roberts:
How does this work with the cache shredding?

We have quite a limited cache space, so adding more keys to the cache doesn't necessarily lead to better build times, but rather the opposite as it forces a rebuild when the cache is thrown out of the window

VinhTruongAmbient:
> How does this work with the cache shredding?
> 
> We have quite a limited cache space, so adding more keys to the cache doesn't necessarily lead to better build times, but rather the opposite as it forces a rebuild when the cache is thrown out of the window

How much cache space do we have?

ten3roberts:
https://github.com/AmbientRun/Ambient/actions/caches

Screenshot 2023-05-31 at 10 33 52

We are using 26 GB out of 10 GB, which means we shred quite a lot

VinhTruongAmbient:
TIL. It looks like we are way above the budget already. Then the only other path is to start sharing caches across different jobs.

ten3roberts:
> TIL. It looks like we are way above the budget already. Then the only other path is to start sharing caches across different jobs.

Yes, though the problem with that is that some builds have different rustc arguments/configuration, so if we share the cache rustc will just invalidate `./target` instead as the build options are incompatible. A common example of this is building with different feature flags (which we do) and running with `#[cfg(test)]` which can force quite deep rebuilds in the tree as conditional modules are now included or excluded and requires rebuilding that crate and its dependents.

Clippy had an issue previously where it didn't work if you ran it after `cargo check`/`cargo build` as clippy is just a reskinned `cargo check` with more checks. It would skip the previously "checked" files so it missed clippy lints, it has now been fixed so that clippy forces a check on previously checked files which were checked with different lints, thought this requires some rebuilding.

VinhTruongAmbient:
Looking at the different cache sizes the Linux one seems particularly large at 4.6GB while others are around 1GB. If all our caches were around 1GB then we can still barely fit within 10GB limit.

ten3roberts:
Locally, running `cargo check` and then `cargo clippy` immediately afterwards yields:

[out.txt](https://github.com/AmbientRun/Ambient/files/11611092/out.txt)

Most importantly:

Screenshot 2023-05-31 at 11 27 57

Which means that if the cache was shared between clippy and build we would recompile our whole local tree in every CI run.

It does not get better when `cargo test` gets thrown into the mix

Regarding the linux cache: It is larger and that is expected since linux does `clippy`, `build`, `test` (generates a *lot* of binaries) and `clippy --target wasm32-unknown-unknown` (different build architecture requires a complete rebuild.

The other operating systems just do the build step.

Cargo test is notoriously for generating large build folders as every integration test and every unit tested crate gets its own binary and linking.

VinhTruongAmbient:
I propose we merge this and create a separate issue for reducing our build sizes to use the cache more effectively. Thoughts?

ten3roberts:
Also

Is there any reason you are using a different caching action than the one we already use?

Does the old one have issues or bugs?

VinhTruongAmbient:
> Also
> 
> Is there any reason you are using a different caching action than the one we already use?
> 
> Does the old one have issues or bugs?

Initially, I thought that they did. Later on, I understood it's the limiting cache factor. Now I just think it's easier to use than what the golden image job did before, which was manually listing all targets. Also, I believe `Swatinem/rust-cache@v2` is more widely used than `Leafwing-Studios/cargo-cache@v1`. It is used by `rust-analyzer` and Rerun folks. I think we can safely switch all our cache implementations to use Swatinem's implementation.

ten3roberts:
I understand.

For my part I am all for switching to that one since it seems more widely used (I have a hunch rust-analyzer is quite used).

However, I think this is an infrastructure decision best suited for @FredrikNoren to decide

FredrikNoren (approved):
@VinhTruongAmbient I'm fine merging this, but just keep an eye on build times; if they go up because of more cache misses then we might want to reconsider

FredrikNoren:
The `Leafwing-Studios/cargo-cache@v` is just there because bevy switched to it, so was hoping that it would be better than the other one. Not sure if it's actually doing anything for us though (compared to the standard one).

Issue #464, opened 2023-05-31
build/assets folder ends up in the wrong place if there are spaces in the path


droqen:
repro: Move the `asset_loading` example into a folder that has spaces in its name, delete the build folder, then cd inside the project and call `ambient build`.

If the folder structure is `Parent Folder/asset_loading` then the built assets will not go to `Parent Folder/asset_loading/build/assets` but instead to `Parent Folder/Parent Folder/asset_loading/assets`

cc @ten3roberts

ten3roberts:
And here I thought we would be safe from these issues using Rust...

ten3roberts:
To clarify, occurs on mac as well.

The parent folder(s) has to contain spaces, not the project name/folder for it to break

ten3roberts:
Having a space inside our guest examples also panics any other ambient run as it results in an invalid workspace configuration due to the stray directory which isn't a rust crates/Cargo.toml.

This specific issue is addressed by #470

PR #465, opened 2023-05-31
Fix spaces in parent project path breaking asset building

ten3roberts:
To clarify, this does not happen if the *project* has a space in it, only if the *parent* folder has a space in it

Fixes: #464


FredrikNoren (approved):
@ten3roberts There's an extra file: [guest/rust/examples/basics/samplers/fail_screenshot.png](https://github.com/AmbientRun/Ambient/pull/465/files#diff-84cb1e15ec837bb18b1e3259c752ea5943b7cc06640c5b79861fc5266f87eea6)

ten3roberts:
> @ten3roberts There's an extra file: [guest/rust/examples/basics/samplers/fail_screenshot.png](https://github.com/AmbientRun/Ambient/pull/465/files#diff-84cb1e15ec837bb18b1e3259c752ea5943b7cc06640c5b79861fc5266f87eea6)

Strange, it seems to appear when I run golden images, haven't touched the sampler example at all.

Occurs on other branches as well, slipped past my git staging

PR #472, opened 2023-06-01
Animation stack transition and playback apis

devjobe:
* Removed AnimationAction::weight - use stack operation to blend instead
* Removed: set_animation_blend (use set_animation_action_stack or play_animation_action_index)
* Added: play_animation_action_index
* Added AnimationActionStack::Transition - transitions the stack to another stack
* Added AnimationActionStack::Start - resets the start time of the action with speed
* Changed: set_animation_action_stack to include transition duration

https://github.com/AmbientRun/Ambient/assets/71590722/f7520cbe-86f1-4371-83bb-495f23fc34e7


devjobe:
Closing this, the api will go back to AnimationActions as its very similar to this approach but perhaps conceptually easier to understand.

Issue #457, opened 2023-05-31
Golden image test: basics/decals fails randomly


VinhTruongAmbient:
Expected image:

![screenshot](https://github.com/AmbientRun/Ambient/assets/129393025/74958a19-d214-4f07-bef7-afd0407967f2)

This test usually passes, but every once in a while this image gets rendered instead:

![fail_screenshot](https://github.com/AmbientRun/Ambient/assets/129393025/3e64967c-0a9b-4b47-93da-0c74d383c24a)

This started happening after #440, but only on Github Actions. On M2 Max the decals example works as expected.

FredrikNoren:
@VinhTruongAmbient Hm that's so strange with the weird grey box in the background in the second screenshot. Seems like it's failing almost every time on main now

VinhTruongAmbient:
I was able to reproduce this on my local Docker container which runs LLVMPipe. Investigating.

VinhTruongAmbient:
As a test I removed all entities, except for the camera and the decal. The result looks like this:

![fail_screenshot](https://github.com/AmbientRun/Ambient/assets/129393025/9436fa3a-aa5e-4331-b699-2ddcd204e748)

The gray box looks to be oriented by the decal, which makes sense. Removing orientation yields:

![fail_screenshot](https://github.com/AmbientRun/Ambient/assets/129393025/183b4749-7544-4a3c-a5ef-2f51e14e30cd)

VinhTruongAmbient:
There seem to be two separate issues:

1. Our decals use `CubeMesh` to draw the context in which the decal is rendered. This `CubeMesh` is actually twice as large in every dimension. The shader expects a cube with `min=[-0.5,-0.5,-0.5], max=[0.5,0.5,0.5]` but our `CubeMesh` is actually `min=[-1.0,-1.0,-1.0], max=[1.0,1.0,1.0]`.
2. `get_decal()` in decal shader calls `discard;` if the projected point does not land inside the unit cube. This is incorrect because in this case the fragment never gets written. So the gray background just happens to be some uninitialized render target data.

Working on a fix.

VinhTruongAmbient:
The uninitialized render target explains the random behavior we have seen in the CI.

VinhTruongAmbient:
Looks like the discard theory doesn't tell the whole story. This bug actually happens on my Mac too, you just had to change the clear color from black to something else:

[...]

PR #473, opened 2023-06-02
Fix decals issue #457

VinhTruongAmbient:
- Decals now use the actual unit cube which has a side length of 1. As a bonus, this might improve decal rendering performance by avoiding unnecessary decal math on pixels outside the unit cube.
- Decals now early outs if the sampled depth equals 0. This avoids the incorrect shading parameters issue where a non-existent normal is sampled from the `screen_normal_quat` render target.
- Updated decals example with a decal texture that is rotationally asymmetric.
- Refactoring decals shader:
  - Simplify shader control flow by avoiding `discard` in the middle of the decal function by using output parameters and returning a bool if the decal test passed.
  - Decal shader no longer outputs `frag_depth`. The decal itself has no geometry and is rendered correctly as long as the projection volume passes the depth test.


FredrikNoren (approved):
Awesome :) Took a quick look but lgtm

PR #479, opened 2023-06-03
add scroll bar

chaosprint:
https://github.com/AmbientRun/Ambient/assets/35621141/9de67ade-1d58-4f24-a088-2b130748663d

### DIFF
--- a/guest/rust/examples/ui/text/src/client.rs
+++ b/guest/rust/examples/ui/text/src/client.rs
@@ -2,7 +2,8 @@ use ambient_api::prelude::*;
 
 #[element_component]
 fn App(_hooks: &mut Hooks) -> Element {
-    FlowColumn::el([
+
+    let e = FlowColumn::el([
         Text::el("Header").header_style(),
         Text::el("Section").section_style(),
         Text::el("Default text \u{f1e2} \u{fb8f}"),
@@ -10,10 +11,27 @@ fn App(_hooks: &mut Hooks) -> Element {
         Separator { vertical: false }.el(),
         Text::el("Custom size").with(font_size(), 40.),
         Text::el("Custom color").with(color(), vec4(1., 0., 0., 1.)),
-        Text::el("Multi\n\nLine"),
+        Text::el("Multi\n\n\nLine"),
+        Text::el("Now").header_style(),
+        Text::el("we").header_style(),
+        Text::el("add").header_style(),
+        Text::el("some").header_style(),
+        Text::el("headers").header_style(),
+        Text::el("to").header_style(),
+        Text::el("test").header_style(),
+        Text::el("the").header_style(),
+        Text::el("scroll").header_style(),
+        Text::el("bar").header_style(),
     ])
     .with_padding_even(STREET)
-    .with(space_between_items(), 10.)
+    .with(space_between_items(), 10.);
+
+    ScrollArea::el(
+        ScrollAreaSizing::FitChildrenWidth,
+        e,
+        600.0,
+        400.0,
+    )
 }
 
 #[main]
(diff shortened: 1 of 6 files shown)


chaosprint:
@FredrikNoren I find it hard to get the height of inner element. Is there a better way or we just leave the height settings manually?

chaosprint:
image

@FredrikNoren I am trying to do things like this:
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y

My idea is to cover it with a Rect with a higher Z axis but this is apparently a hack. Is there some better way to do it?

FredrikNoren:
@chaosprint Hm yeah covering it won't really work. We need to do clipping on all child-elements. So basically something like this:

- We add a `clip: Rect` component, which specifies a left, right, width and height for the clipped region in world space. You set this to the root entity of a scrollbar
- Then that component needs to automatically propagate to all child entities, including changes to it.
- Then we need to implement something in the renderer which takes the clipping into account, so that anything that is outside the clipped region is hidden

guest/rust/examples/ui/text/src/client.rs
```diff
@@ -2,18 +2,39 @@ use ambient_api::prelude::*;
...
+    let e = FlowColumn::el([
         Text::el("Header").header_style(),
         Text::el("Section").section_style(),
         Text::el("Default text \u{f1e2} \u{fb8f}"),
         Text::el("Small").small_style(),
         Separator { vertical: false }.el(),
         Text::el("Custom size").with(font_size(), 40.),
         Text::el("Custom color").with(color(), vec4(1., 0., 0., 1.)),
-        Text::el("Multi\n\nLine"),
+        Text::el("Multi\n\n\nLine"),
```
FredrikNoren, line 14: Let's just do a separate example for scrollbars, and keep the text one for text

shared_crates/ui/src/scroll_area.rs
```diff
@@ -23,6 +27,8 @@ pub enum ScrollAreaSizing {
     FitChildrenWidth,
     /// Assumes the width from the parent and propagates it to the children
     FitParentWidth,
+    /// Fit to children width, height is screen and assign a max scroll down
```
FredrikNoren, line 30: Hm I don't think this makes sense as an option to this enum, it's got nothing to do with how the width is adjusted.

shared_crates/ui/src/scroll_area.rs
```diff
@@ -32,14 +38,29 @@ pub fn ScrollArea(
...
     sizing: ScrollAreaSizing,
     /// The child element
-    inner: Element,
+    inner: Element
 ) -> Element {
     let (scroll, set_scroll) = hooks.use_state(0.);
     let (inner_size, set_inner_size) = hooks.use_state(Vec2::ZERO);
-    hooks.use_runtime_message::<messages::WindowMouseWheel>(move |_world, event| {
+    let res = use_window_logical_resolution(hooks);
+    let max_scroll_down = match sizing {
```
FredrikNoren, line 46: There's a MeasureSize component we should use to figure out what the size of the ScrollArea is, instead of supplying it from outside

chaosprint:
> @chaosprint Hm yeah covering it won't really work. We need to do clipping on all child-elements. So basically something like this:
> 
> * We add a `clip: Rect` component, which specifies a left, right, width and height for the clipped region in world space. You set this to the root entity of a scrollbar
> * Then that component needs to automatically propagate to all child entities, including changes to it.
> * Then we need to implement something in the renderer which takes the clipping into account, so that anything that is outside the clipped region is hidden

So where should I put this `clip: Rect` component?

PR #482, opened 2023-06-03
Fixed mutate_component documentation

aldzban:
The documentation for `mutate_component` had a default `name` instead of the name of the parameter `component`.


FredrikNoren:
Thanks!

PR #485, opened 2023-06-05
feat: async buffered golden image test

ten3roberts:
- Non blocking test running with automatic concurrency
- Display of in-flight tests currently runnning
- Reduced contention on build directory lock file
- Spinner for single test rather than "stuck" (0/1) bar.
- Probably some more that I forgot to mention


ten3roberts:
Screenshot 2023-06-05 at 16 23 52

PR #487, opened 2023-06-06
add scroll bar to the `ScrollArea` element and add a `scroll` example

chaosprint:
https://github.com/AmbientRun/Ambient/assets/35621141/d7fc1865-f655-42eb-bf72-820dac7c5166


guest/rust/examples/ui/scroll/src/client.rs
```diff
@@ -0,0 +1,78 @@
+use ambient_api::prelude::*;
+
+#[element_component]
+fn App(hooks: &mut Hooks) -> Element {
+    let size_info = hooks.use_query(window_logical_size());
+    let x = size_info[0].1.x as f32;
+    let y = size_info[0].1.y as f32;
+    let f = FlowColumn::el([
```
FredrikNoren, line 8: Use the `WindowSized` component instead of reading out the window size manually here

shared_crates/ui/src/scroll_area.rs
```diff
@@ -68,3 +72,111 @@ pub fn ScrollArea(
         }
     }
 }
+
+
+/// A scroll box view that can be used to scroll its child with a scroll bar.
+#[element_component]
+pub fn ScrollBoxView(
```
FredrikNoren, line 79: We should merge the ScrollArea and ScrollBoxView, I don't see any reason both should exist?

shared_crates/ui/src/scroll_area.rs
```diff
@@ -68,3 +72,111 @@ pub fn ScrollArea(
...
+    hooks: &mut Hooks,
+    /// Width of the scroll area
+    min_width: f32,
+    /// Height of the scroll area
+    min_height: f32,
+    /// Distance of scrolling in y axis
+    scroll_height: f32,
```
FredrikNoren, line 86: Scroll height should be automatically measured internally. Use the `MeasureSize` component on the inner content. You can also see how it's done in the ScrollArea component

shared_crates/ui/src/scroll_area.rs
```diff
@@ -68,3 +72,111 @@ pub fn ScrollArea(
         }
     }
 }
+
+
+/// A scroll box view that can be used to scroll its child with a scroll bar.
+#[element_component]
+pub fn ScrollBoxView(
+    hooks: &mut Hooks,
+    /// Width of the scroll area
```
FredrikNoren, line 81: width and height should be automatically calculated through the layout, not passed as parameters

FredrikNoren (changes requested):
Nice! Looks like a good start. Added some comments. The big thing is to make this work with the layout system, which I think the ScrollArea already does, so might make sense to just add the scissors and thumb to that one

shared_crates/ui/src/scroll_area.rs
chaosprint: The `MesureSize` does not return a proper height of a `Flow`, is this something that should be fixed first?

FredrikNoren: Hm that sounds fishy, it's a very simple component. Check the source of it

shared_crates/ui/src/scroll_area.rs
chaosprint: The `BoxViewHeight` has to be specified. The inner height can be very large.

FredrikNoren: So there are two heights; inner and outer. Both should be automatically calculated by the layout system. The outer one the layout system will calculate based on how the user uses the scroll area (if they slot it in somewhere it should automatically calculate the size for it). The inner is also calculated by the layout system, but based on the content of the inner elements.

chaosprint: > 

https://github.com/AmbientRun/Ambient/assets/35621141/2322c44a-c723-4470-953d-45279945f494

But how can an outer height be calculated? It doesn't make sense at all to me. Take the small area above as an example: the height can be very small, small enough to only accommodate two lines of numbers, or it can be very large, large enough to accommodate eight lines of numbers. This value should be user-specified to determine the area's height and also the `scissors` height.

The inner height should be calculated I totally agree. In that case, the inner height should be 9 lines of numbers.

guest/rust/examples/ui/scroll/src/client.rs
chaosprint: Can you give an example to get the size of `WindowSized`?
I can manage to get the ID of `WindowSized` but it's `width()` or `height()` does not exist unless manually specified.

shared_crates/ui/src/scroll_area.rs
chaosprint: > We should merge the ScrollArea and ScrollBoxView, I don't see any reason both should exist?

After a series of frustrating attempts, I've come to appreciate the importance of a new API. The current naming of `ScrollArea` is misleading, as the sizing of this `ScrollArea` allows users to choose to fit the parent or children's width, and its movement is vertical, so it should actually be called 'VerticalScrollArea'. Also, adding a `Direction` attribute to it is not the most ergonomic approach. It shares a common issue with our `components`: lacking in coherence; if a user uses `Direction::Vertical` but is allowed to freely choose to fit height or width in the sizing entry, it's counterintuitive and error-prone. A better approach would be to rename it to `VerticalScrollView` and add `HorizontalScrollView` when needed.

In short, I don't think it's a good idea to cram everything into a single `ScrollArea` as it results in an excess of internal settings and illogical pairing. We also have FlowColumn and FlowRow, so we should maintain logical coherence here.

This is why I want to start with `ScrollBoxView`, allowing users to specify all widths and heights. Rather than merging it into ScrollArea, I think it's more important to provide units similar to 'em' and 'vh' in CSS (currently, everything is in px). However, I agree that the overflow height and width should be calculated automatically. My thought on height calculation can be seen below as well.

FredrikNoren (approved):

PR #493, opened 2023-06-11
De-dupe bind ids when they use the same name

### LINKED ISSUE #419: Blender Animation transforms are off on Ambient import
I have a rigged and animated character in Blender. I'm exporting as gltf/glb. The pose keyframes are getting transformed weirdly when the get to ambient.

Here's an example of the 'idle' pose. Which is one keyframe

## Idle
* Blender
image

* Ambient
Screen Shot 2023-05-23 at 2 05 20 PM

You can notice the shoulders and knee joints are kinked and off and the hitbox is not bound to the hand.

## Walk
* Blender
walk

* Ambient
walkAmbient

How can I further troubleshoot this? I just redid the rig and animations and maybe missed baking some transforms.

you can find the code at https://github.com/kevzettler/ambient-test-project/tree/new_animations on the `new_animations` branch

FredrikNoren:
Fixes #419


kevzettler:
@FredrikNoren I just tested this and confirms it fixes all my issues with #419 thank you for hunting this down!

Issue #461, opened 2023-05-31
Get ambient to build without protoc


FredrikNoren:
We shouldn't require protoc to be installed to build ambient. I suspect we can run it "offline" and check in the generated code instead? (I know that right now it's hidden behind a feature flag, but pretty soon we'll want it to always be enabled)

FredrikNoren:
Rust-analyzer is currently failing on main if protoc is not installed.

PR #494, opened 2023-06-12
Add generated proto file to stop relying on protoc

pierd:
https://github.com/AmbientRun/Ambient/issues/461


FredrikNoren (approved):
Nice, maybe we should just have either a doc or a sub-command for campfire for updating the proto? (But fine to just merge as is for me)

pierd:
> Nice, maybe we should just have either a doc or a sub-command for campfire for updating the proto? (But fine to just merge as is for me)

Cloud repo generates it (in its `build.rs`). Updated the README.

Issue #375, opened 2023-05-08
Broken IDE intellisense for input, frame modules and others.


kevzettler:
I'm working on some input handling features and have some Ambient code like:

```
   let mut cursor_lock = input::CursorLockGuard::new(true);
    ambient_api::messages::Frame::subscribe(move |_| {
        let (delta, input) = input::get_delta();
```

I wanted to learn more about `input::get_delta` and `messages::Frame`  However my IDE cannot look up any definitions,implementations, or declarations for these modules. I have tried in both emacs, and vscode There are gifs below.

in emacs:

![emacs](https://user-images.githubusercontent.com/126209/236883211-aa63f4df-31e2-4454-853d-a8b9ab6ea54c.gif)

And in Vscode

![vscode](https://user-images.githubusercontent.com/126209/236883251-48259ac3-04c7-45fd-aaaa-5ff33369ecd1.gif)

 Is this a greater configuration issue on my machine? If this is expected behavior this is very unfortunate developer experience. Seems to break expected conventions and very cumbersome to have to go look at Ambient source code directly.

philpax:
This is definitely not intentional; it's an issue with rust-analyzer not being able to handle the attribute proc macros. If you comment out the `#[main]` attribute, your intellisense will work properly.

I'll see if I can chat to the rust-analyzer folk about why this is happening, and what we can do to prevent it.

kevzettler:
yes I do see the `#[main]` macro highlighted with:

``` server.rs    51   1 error    unresolved-proc-macro proc macro `main` not expanded: no proc macro dylib present (lsp)```

@philpax this is currently happening in my public project at:
https://github.com/kevzettler/ambient-test-project

**Comment by philpax** [comment]
Cheers, I'll have a look soon!

I notice that you have a direct dependency on `ambient_api`:
```toml
ambient_api = { path = "/Users/kevzettler/code/Ambient/guest/rust/api" }
```

Do you still see the same issue if you switch to using `ambient_api = "0.2.1"`?

kevzettler:
@philpax I just switched to `ambient_api = "0.2.1"` and did `cargo update`

```
/Users/kevzettler/code/ambient-test-project
kevzettler@kevs-mbp-3 ambient-test-project % cargo update
    Updating crates.io index
      Adding ambient_api v0.2.1
      Adding ambient_api_core v0.2.1
      Adding ambient_api_macros v0.2.1
      Adding ambient_cb v0.2.1
[...]

PR #505, opened 2023-06-15
Add Emacs section to IDE setup docs

kevzettler:
addresses #375


ten3roberts:
Looks splendid. I will have to add instructions for Neovim as well now.

PR #511, opened 2023-06-16
Revamp deployments

pierd:
Ended up being an almost complete rewrite to support:
- big files - previously only <4MB
- differential uploads - so that assets that haven't been changed don't have to be reuploaded


app/src/main.rs
```diff
@@ -284,16 +287,17 @@ fn main() -> anyhow::Result<()> {
             anyhow::bail!("Can only deploy a local project");
         };
         let manifest = manifest.as_ref().expect("no manifest");
-        let response = runtime.block_on(ambient_deploy::deploy(
```
ten3roberts, line 287: My upcoming PR #444 will allow `.await` here rather than sprinkled `block_on`s

crates/deploy/proto/ambient.run.deploy.rs
ten3roberts, line 1: I suppose this file is auto generated.

If so we should consider adding it to our docs similar to what we have done to the `bindings.wit` file

crates/deploy/src/lib.rs
```diff
@@ -2,51 +2,174 @@ pub mod deploy_proto {
...
 
-use std::{path::Path, str::FromStr};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use ambient_project::Manifest;
+use md5::Digest;
```
ten3roberts, line 12: Why md5 rather than sha256?

I may be wrong but isn't MD5 considered cryptographically insecure as collisions can be fabricated using normal computers, which would set us up for DoS, though I may just be concerned in vain.

crates/deploy/src/lib.rs
```diff
@@ -2,51 +2,174 @@ pub mod deploy_proto {
...
-    base_path: impl AsRef<Path>,
+async fn asset_requests_from_file_path(
+    ember_id: impl AsRef<str>,
+    asset_path: impl AsRef<str>,
     file_path: impl AsRef<Path>,
-) -> anyhow::Result<DeployAssetRequest> {
-    let path = file_path
-        .as_ref()
-        .strip_prefix(base_path)?
-        .to_string_lossy()
```
ten3roberts, line 25: I think failing here if a file path isn't UTF-8 would be preferred, since this can map different files to the same value, which can be a very nasty bug to figure out, especially for users who don't know this happens, as they think they did some configuring wrong.

crates/deploy/src/lib.rs
```diff
@@ -83,39 +206,110 @@ pub async fn deploy(
...
+}
 
-    // iterate over all files to deploy (everything in the build directory)
-    let file_paths = WalkDir::new(path.as_ref().join("build"))
-        .into_iter()
-        .filter_map(|e| e.ok())
-        .filter(|e| e.metadata().map(|x| x.is_file()).unwrap_or(false))
-        .map(|x| x.into_path());
+enum FileRequest {
+    SendMD5(String),
```
ten3roberts, line 221: Same here regarding UTF-8.

I would prefer to use the appropriate type for the job, unless this is an encoded path, (in which case a newtype may be more clear)

ten3roberts (changes requested):

crates/deploy/src/lib.rs
FredrikNoren: @ten3roberts You need to consider where it's used as well; this is for hashing asset files for fingerprints, not for cryptography.

crates/deploy/Cargo.toml
```diff
@@ -22,3 +22,4 @@ walkdir = { workspace = true }
 prost = "0.11"
```
FredrikNoren, line 22: Let's put all dependencies in the root Cargo.toml (and use {workspace=true}), otherwise we'll never be able to de-dupe them

FredrikNoren (approved):
Nice!

crates/deploy/src/lib.rs
ten3roberts: Of course, though wouldn't a man in the middle be possible since you can send a different fabricated file that has the same MD5 sum, or mischievously replace things.

Do I think that is not a large concern of us as we don't use it for security related functionality or banking.

I just wanted to shed light on the possibility just so we at least had a discussion and evaluation of it.

pierd: MD5 mostly because it's already used by Google Cloud Storage that we use in the backend. It calculates it for us when a file is uploaded.

As for collisions it shouldn't be a problem (basically what @FredrikNoren said).

FredrikNoren: @ten3roberts Same as the discussion in the channel; we need to prioritize what we put our energy into. Right now three of us are putting time and energy into discussing this problem; which means we're not working on something more important. It's important to make sure we only start discussions that are actually important, otherwise we drown in the noise (if we just discuss everything). The problem we're discussing here is clearly extremely hypothetical, and if we listed all of the things we need to do to make ambient work it would come extremely far down; so we clearly should not put time and effort into working on it (by discussing it).

crates/deploy/src/lib.rs
pierd: Agreed, better to fail early. I'll update it.

crates/deploy/proto/ambient.run.deploy.rs
pierd: It's generated in an internal repo and added here to not depend on having protoc available in the system (https://github.com/AmbientRun/Ambient/issues/461).

PR #513, opened 2023-06-20
add resizing to debugger

chaosprint:


FredrikNoren (approved):
Nice! In the future it would be nice to pull it out into a component (like `Resizable` or something).

PR #522, opened 2023-06-21
Port systems and rendering logic to web client

ten3roberts:


web/client/src/app.rs
ten3roberts, line 1: Note:

Code inside `web/` is more commented out as it is very much work in progress.

This code could be removed and referred to in git history, but such a solution makes it difficult to find the pieces in history as well as remember to uncomment them when the time arises.

FredrikNoren (approved):

PR #528, opened 2023-06-22
fix: replace decals material with less uniform texture

ten3roberts:


ten3roberts:
It seems that the image differ is desensitivized on grid like feature

Issue #191, opened 2023-03-01
Unable to run examples on old GPU


Kaesebrot84:
My GPU does not support vulkan so I was trying to run the "tictactoe" example (0.1.1) with the `gl` backend by exporting the according environment variable:

```
export WGPU_BACKEND=gl
```
Setting the backend does not seem to work in this case:

```
[2023-03-01T16:36:39Z INFO  ambient] Building Tic Tac Toe
[2023-03-01T16:36:40Z INFO  ambient] Done building Tic Tac Toe
[2023-03-01T16:36:40Z INFO  ambient::server] Creating server
[2023-03-01T16:36:40Z INFO  ambient::server] Created server, running at 192.168.178.86:9000
[2023-03-01T16:36:40Z INFO  ambient::server] Starting server
/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/physx-sys-0.8.2/PhysX/physx/source/physx/src/NpScene.cpp (2159) : invalid operation : PxScene::fetchResults: fetchResults() called illegally! It must be called after advance() or simulate()

[2023-03-01T16:36:40Z INFO  winit::platform_impl::platform::x11::window] Guessed window scale factor: 1
[2023-03-01T16:36:40Z ERROR wgpu_hal::vulkan::instance] enumerate_adapters: Initialization of an object has failed
[2023-03-01T16:36:40Z ERROR wgpu_hal::vulkan::instance] enumerate_adapters: Initialization of an object has failed
thread 'main' panicked at 'Failed to find an appropiate adapter', /home/runner/work/Ambient/Ambient/crates/gpu/src/gpu.rs:55:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'A Tokio 1.x context was found, but it is being shutdown.', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.25.0/src/runtime/time/entry.rs:567:13

```

Just for comparison I built the [bevy breakout example](https://github.com/bevyengine/bevy#getting-started) with the same environment variable and it set the backend correctly.

```
2023-03-01T16:43:06.878129Z  INFO bevy_winit::system: Creating new window "Bevy App" (0v0)
2023-03-01T16:43:06.878669Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1    
[...]

PR #535, opened 2023-06-27
Fix issues with web rendering and debugger

ten3roberts:


ten3roberts:
Same example:

image

ten3roberts:
AU tests fails on master as the examples was moved

ten3roberts:
They work(ed) locally

ten3roberts:
This PR contains a lot of changes which further brings the web client experience forwards and makes it more reliable, consistent and correct.

Some of the things worthwhile to mention:

> For a full list of all the things I've made sure that the commit messages are relevant and some for larger/trickier bugs contain a longer description. Please have a look at the full log for a complete list.

The renderer contains many adjacent arrays which are both one dimensional, and multidimensional. This ensures that there arrays have their relationship being consistent and at all times well defined, without causing small buffer overwrites into another sub buffer etc.

Built out CPU side indirect draw command generation to exclude the culling and get a consistent CPU debugging of the draw commands, as well as removing the unpredictable order of the commands from run to run depending on the compute shaders. For future debugging or older machines it is accessible through our GPU settings (`~/Library/Application Support/com.Ambient.Ambient/settings.toml`)

Add a fallback direct rendering mode to diagnose indirect issues by using direct draw calls.

The buffer allocation on wgpu overcommits and is thus nicer to us and allows one or two elements to be written out of bounds before hitting an actual out of bounds error. This essentially hides many off by one errors, especially in the *multi*-buffer where you would overwrite another subbuffer.

Found a plethora of issues regarding buffer overreads, overwrites, as well as incorrect bounding ranges on both the CPU and GPU. Additionally, the length was in many places mixed up with capacity. This meant that the buffer was written and used with data that was more than its length, but within the capacity (or the capacity + wgpu's offset adjustements)

For the Typed buffer, the length/count was in places mixed up with the byte size. For example: doing `read(..)` did not read the whole buffer, but read "count" *bytes* instead of "count" *items*.

Another issue with the typed buffer is that it would be written to by the GPU, but the `len` never got updated to reflect this data, which caused the buffer to appear shorter than it actually was. 

The read back of the culling results could (and did) arrive out of order which means that frame #3 culling data could arrive after frame #4 causing you to render data which is older than you already know. This became even worse during high tokio runtime load due to pre-emptive cooperative scheduling. This is normally not a larger problem than having an object flicker in and out out of nowhere as a frame from a few seconds ago arrived. However, if a material was removed during the time, or primitives where added and the layouts where consequently shift then the rendering will *cause an out or bounds read*.

A silent *undocumented* no-op occurs on the web if the `indexed-first-instance` feature is not supplied during device creation. This causes only the first global primitive to be rendered, while all others silently no-op without a warning or error in the console.

Overcommit in the compute shader caused primitive fighting which could be observed as occasional flickering even if *no* object left the viewport

In general, the reason for all these changes is that *web*gpu is a lot more strict and less permissive than *w*gpu, for example buffer usage and overwrites. There have also been a bunch of race conditions and fighting which violated GPU memory safety, which weren't always caught on native since our buffers were large enough

crates/gpu/src/gpu.rs
```diff
@@ -84,11 +86,16 @@ impl Gpu {
...
-        let features =
-            wgpu::Features::MULTI_DRAW_INDIRECT | wgpu::Features::MULTI_DRAW_INDIRECT_COUNT;
+        cfg_if::cfg_if! {
+            if #[cfg(target_os = "macos")] {
+                let features = wgpu::Features::empty();
+            } else if #[cfg(target_os = "unknown")] {
+                let features = wgpu::Features::INDIRECT_FIRST_INSTANCE;
+            } else {
+                let features =
+                wgpu::Features::MULTI_DRAW_INDIRECT | wgpu::Features::MULTI_DRAW_INDIRECT_COUNT;
```
philpax, line 96: Maybe add a comment here explaining how the renderer works in each of these cases, so that someone reading the code knows what tradeoffs are present for each architecture

crates/gpu/src/multi_buffer.rs
```diff
@@ -201,51 +226,68 @@ impl MultiBuffer {
...
-        // Wgpu requires copy_buffer_to_buffer to align to COPY_BUFFER_ALIGNMENT, so each of our sub-buffers need to align to that too
+        // wgpu requires copy_buffer_to_buffer to align to COPY_BUFFER_ALIGNMENT, so each of our sub-buffers need to align to that too
         assert_eq!(capacity % wgpu::COPY_BUFFER_ALIGNMENT, 0);
         let buf = self.sub_buffers[id].unwrap();
+
         let capacity_change = capacity as i64 - buf.capacity_bytes as i64;
         let new_total_capacity = (self.total_capacity as i64 + capacity_change) as u64;
+        tracing::debug!(
+            label = self.label,
+            "Rezing sub buffer {id} {} => {}",
```
philpax, line 261: Is this "resizing" or "reviving"?

crates/gpu/src/multi_buffer.rs
```diff
@@ -410,6 +481,17 @@ impl<T: bytemuck::Pod> TypedMultiBuffer<T> {
         offset: u64,
         data: &[T],
     ) -> Result<(), MultiBufferError> {
+        // TODO: correctly update len for the written subbuffer
```
philpax, line 484: Does this need to be dealt with for the merge?

crates/gpu/src/gpu.rs
ten3roberts: Wanted to keep this PR "small" since there is an infinite space of possibilities we could unlock.

For example at runtime configuring this feature to be off on windows as well (#191) etc.

Will do though

crates/gpu/src/multi_buffer.rs
ten3roberts: Resizing. Good

crates/gpu/src/gpu.rs
philpax: Yeah no stress, a short explanation is sufficient

crates/renderer/src/tree_renderer.rs
```diff
@@ -217,53 +250,164 @@ impl TreeRenderer {
...
-            &self.primitives,
-            collect_state,
-            self.primitives.total_len() as u32,
-            material_layouts,
-        );
+        assert_eq!(mem::size_of::<DrawIndexedIndirect>(), 20);
+
+        assert_eq!(mem::align_of::<DrawIndexedIndirect>(), 4);
+
+        if self.config.render_mode == RenderMode::Direct {
```
philpax, line 293: Is this meant to be empty?

guest/rust/examples/basics/physics/Cargo.toml
```diff
@@ -9,10 +9,10 @@ version = {workspace = true}
...
 ambient_api = {workspace = true}
 
-[[bin]]
-name = "physics_client"
-path = "src/client.rs"
-required-features = ["client"]
+# [[bin]]
+# name = "physics_client"
+# path = "src/client.rs"
+# required-features = ["client"]
```
philpax, line 15: Shouldn't be commented out; I'd suggest just disabling WASM loading entirely on the client for now

philpax (approved):
Honestly, this is too large for me to understand in its totality. However, the individual bits look good. Excellent work - if it works across all the platforms and there's no other cleanup work to be done, this is good to go!

crates/renderer/src/tree_renderer.rs
ten3roberts: Yes. The alternative is `if software_culling && render_mode == Indirect {} else if !software_culling && render_mode != Direct`

This makes for less inversions and duplicate boolean expressions

crates/gpu/src/multi_buffer.rs
ten3roberts: I couldn't do it as well since only `&self` is available, and this would need `&mut self`.

ten3roberts:
I've tested it on my windows laptop as well, so it works on all our supported platforms

philpax (approved):

PR #550, opened 2023-06-29
feat(api): add entity::get_[all_]components

### LINKED ISSUE #90: Add `entity::get_all_components`
You should be able to find all the components attached to an entity. The reason I haven't done this yet is because it's a little complicated as a script is not actually guaranteed to know about all of the components in the world, but it should be possible with a dict from id to known components. (Or we just force people to pull components out of the `Components` one by one. That would work today, but it wouldn't be very useful for entity introspection.)

### LINKED ISSUE #220: Implement Debug on Entity (guest API)
Would be great with if `Entity` implemented `std::fmt::Debug`:  https://github.com/AmbientRun/Ambient/blob/main/guest/rust/api/src/internal/component/mod.rs#L56 

It should print each component and its value.

### LINKED ISSUE #404: `entity::get_components(id, tuple)`
The concept functions return a tuple of components that could be potentially used to retrieve multiple components from an entity at the same time:

```rust
let (comp1, comp2) = entity::get_components(id, concepts::cool_concept())?;
```

I don't think this should be too conceptually difficult to implement - it's pretty similar to what `GeneralQuery` already does.

### LINKED ISSUE #517: Allow the retrieving of an entire entity by ID
I'd like to be able to clone entire entities, including components that the cloning code may not be aware of. The purpose is to be able to instantiate entity [prototypes](https://refactoring.guru/design-patterns/prototype) that contain mod-defined components relating to mod-defined behavior.

@philpax mentioned adding an `entity::get_all_components` guest API call that could do this, and this definitely works for my use case, especially since I can look up non-mod components before they are spawned into the world and trigger spawn systems.

This is pretty high-priority because this design pattern will be very common in Flowerpot.

philpax:
Closes #220, closes #404, closes #90, closes #517.

- Rewrites the ECS manipulation code to shunt all of the conversion into two functions, so that the operations can be written separately from the conversion.
- Adds `entity::get_components`
- Adds `entity::get_all_components`
- Implements `Debug` for `Entity` on the guest


marceline-cramer (commented):
I'm not very familiar with Ambient's low-level component manipulation and Wasm API code, but I do like the revised `define_component_types!()` macro and how it moves macro-specific code out into multiple functions that can use Rust's type system to its advantage.

PR #552, opened 2023-06-30
Split `time` into `absolute_game_time` and `absolute_epoch_time`

philpax:
The use of the UNIX epoch for the existing `time()` function resulted in `as_secs_f32` on the guest suffering from integer rounding of f32s (as the current Unix timestamp is past 2^24-1). To avoid this, I've split `time` into `absolute_game_time` and `absolute_epoch_time`, so that users can know what to expect and to pick the one that's more appropriate for their application.

Along the way, I made these other fixes: 

- Be more consistent about the use of `Instant` vs `SystemTime` for ticking
- Use a fixed server tick across the board
- Unify time handling logic across client and server


ten3roberts (approved):
👍

PR #553, opened 2023-06-30
Add instancing example

ten3roberts:


app/src/client/mod.rs
```diff
@@ -456,7 +456,7 @@ fn GameView(hooks: &mut Hooks, show_debug: bool) -> Element {
                 }),
             }
             .el()
-            .with(docking(), ambient_layout::Docking::Bottom)
+            .with(docking(), ambient_layout::Docking::Top)
```
ten3roberts, line 459: This makes the debugger dropdown accessible as they don't go off the bottom of the screen

philpax (approved):

philpax:
I'm going to merge this once the checks pass, but something worth noting - updating the entities from the server results in 6MB/s (!) of bandwidth - we should consider reducing the size and/or moving this to the clientside once that's available on the web.

PR #556, opened 2023-07-03
Rename instancing -> ambient_example_instancing

pierd:
... to match other examples

### DIFF
--- a/guest/rust/Cargo.lock
+++ b/guest/rust/Cargo.lock
@@ -252,6 +252,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "ambient_example_instancing"
+version = "0.3.0-dev"
+dependencies = [
+ "ambient_api",
+]
+
+[[package]]
 name = "ambient_example_messaging"
 version = "0.3.0-dev"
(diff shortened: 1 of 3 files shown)


ten3roberts:
Great.

I just used `ambient new ./guest/rust/examples/basic/instancing` to create it.

Is there another preferred way to create our own examples that I've missed?

ten3roberts (approved):
Works on my end and branch 👍

pierd:
I don't think that we have a process for adding examples (yet). I'm assuming that the directory names for examples are shortened just like for our crates (`ambient_network` crate in `network` dir).

I've only noticed because I've added a GitHub action for deploying all examples weekly.

pierd:
BTW this example is quite trippy 😵‍💫 🤣

ten3roberts:
I've discovered an overrun write in our culling and collect compute shader that *only* occurs on the web, so this is a smaller reproduction of it. Discovered a quite massive overcommit in the culling.

ten3roberts:
The samplers example have failed locally for me before. Hopefully it is just flaky

PR #561, opened 2023-07-04
Fix race condition between connection establishment, diff streaming, and game client setup

ten3roberts:


ten3roberts:
This will also fix the spurious *golden-images* we have seen on master and locally recently

app/src/server/mod.rs
```diff
@@ -310,11 +310,18 @@ fn start_http_interface(
...
     };
 
+    let project_path = project_path.map(ToOwned::to_owned);
+
     runtime.spawn(async move {
         let addr = SocketAddr::from(([0, 0, 0, 0], http_interface_port));
 
-        if let Err(err) = serve(addr).await {
-            tracing::error!("Failed to start server on: {addr}\n\n{err:?}");
+        tracing::info!(?project_path, "Starting HTTP interface on: {addr}");
```
ten3roberts, line 318: You know me and `tracing` by now 😛

I reckon this one is actually *useful* for users, but we may demote it to a debug!

crates/network/src/client.rs
```diff
@@ -63,7 +63,7 @@ pub type DatagramHandlers = HashMap<u32, (&'static str, DatagramHandler)>;
 /// Represents either side of a high level connection to a game client of some sort.
 ///
 /// Allows making requests and RPC, etc
-pub trait ClientConnection: 'static + Send + Sync {
+pub trait NetworkTransport: 'static + Send + Sync {
```
ten3roberts, line 66: Renamed `ClientConnection` to `NetworkTransport`.

This is used by both the server and the client to abstract over a kind of network connection. It abstracts over quinn, but also a proxy for the webtransport driver. `WebTransportSession` is not send, and therefore the message queue proxy has to be used instead. Hence, it isn't really a *Connection* trait, but rather a *Transport* for a connection. Besides, *Connection* is already imported in scope in many files, most commontly `quinn::Connection`, which is wrapped by our ambient proxy `ConnectionKind` and *then* becomes a transport.

This makes the naming a lot clearer and avoids using *Client...* inappropriately, and also avoids just remixing the same words into different types.

ten3roberts:
It is a *lot* more robust locally now, though there still seems to be a few timing related issues that fail in the CI. Have only managed to reproduce it once locally, but at least it panics directly rather than failing and tumbling down 😄 

https://github.com/AmbientRun/Ambient/actions/runs/5473029040/jobs/9966383835#step:8:343

campfire/src/golden_images/mod.rs
ten3roberts, line 1: This makes it easier to see which tests fail by printing the names at the bottom, rather than finding the failing test in the screenfulls of backtraces

campfire/src/golden_images/mod.rs
```diff
@@ -284,8 +302,14 @@ async fn run<S: AsRef<str>>(
 
     if !failures.is_empty() {
         for failure in &failures {
-            failure.log();
+            eprintln!("{failure}")
```
ten3roberts, line 305: This is not a log but part of the program output

app/src/server/mod.rs
philpax: Not sure if we're letting users know about the HTTP interface, but it might be useful for people who want to portforward or something 🤔

app/src/server/mod.rs
```diff
@@ -310,11 +317,18 @@ fn start_http_interface(
...
+
+        if let Err(err) = serve(addr)
+            .await
+            .with_context(|| format!("Failed to start server on: {addr}"))
+        {
+            tracing::error!("{err:?}");
```
philpax, line 331: Should this exit if it fails?

crates/network/src/native/client.rs
```diff
@@ -72,7 +76,7 @@ pub struct GameClientView {
     pub inner: Element,
 }
 
-impl ElementComponent for GameClientView {
+impl ElementComponent for ClientView {
```
philpax, line 79: I wonder if this can be transitioned to a `#[element_component]`?

crates/network/src/native/client.rs
```diff
@@ -119,33 +110,32 @@ impl ElementComponent for GameClientView {
                 control_tx.send(Control::Disconnect).ok();
             }
         });
+        let gpu2 = Arc::clone(&gpu);
```
philpax, line 113: Why clone here and not in the scope of the places it's used? It wasn't immediately obvious to me where/why you'd be cloning this.

crates/network/src/native/client.rs
```diff
@@ -119,33 +110,32 @@ impl ElementComponent for GameClientView {
...
             let world_event_reader = Mutex::new(hooks.world.resource(world_events()).reader());
 
-            let game_client_exists = game_client.is_some();
+            let client_state = client_state.clone();
             hooks.use_frame(move |app_world| {
-                if !game_client_exists {
-                    return;
-                }
+                if let Some(client_state) = &client_state {
+                    let mut game_state = client_state.game_state.lock();
```
philpax, line 123: Will anything be trying to lock this at the same time?

crates/network/src/native/client.rs
```diff
@@ -155,42 +145,68 @@ impl ElementComponent for GameClientView {
 
         let (err, set_error) = hooks.use_state(None);
 
-        hooks.use_task(move |_| {
+        let assets2 = assets.clone();
```
philpax, line 148: ditto on inside-clone

crates/network/src/web/client.rs
```diff
@@ -63,77 +67,102 @@ impl ElementComponent for GameClientView {
...
-        hooks.use_local_task(|_| {
+        hooks.use_local_task(move |ui_world| {
+            let local_resources = world_instance_resources(AppResources::from_world(ui_world))
+                .with(game_screen_render_target(), render_target.0.clone());
+
             let task = async move {
+                // NOTE: this is here to simulate a little connection delay to test the
+                // robustness for execution before connect etc
+                //
+                // TODO: remove once the web is fully working
```
philpax, line 101: I suspect this might be the kind of thing we forget about; maybe add a log here to indicate that it slept?

crates/std/src/uncategorized/asset_url.rs
```diff
@@ -236,7 +236,10 @@ impl AbsAssetUrl {
         Ok(Self(self.to_download_raw_url(assets)?))
     }
     fn to_download_raw_url(&self, assets: &AssetCache) -> Result<Url, url::ParseError> {
-        self.to_download_url_with_base(&ContentBaseUrlKey.get(assets))
+        let content_url = ContentBaseUrlKey.get(assets);
+        log::info!("Converting {self} to a download url. Found {content_url:?}");
```
philpax, line 240: Log necessary?

philpax (approved):
Awesome work! Looks much clearer / less race-y now 🙂 

Just a few comments, but I think this is otherwise good to go. I trust you've tested this on both desktop and web and haven't encountered any issues.

app/src/server/mod.rs
ten3roberts: I think it should, I just kept the current behavior at this PR was mostly for fixing issues, rather than changing existing behavior.

app/src/server/mod.rs
ten3roberts: Yes, though I suspect that we will aim towards people not having to portforward as our goal is a "share a link and play" experience.

I have way to many memories of me port forwarding Minecraft servers, as with the error handling is it the same thing: ambient should do the background work as much as possible so that users don't even have to think about it.

crates/network/src/native/client.rs
ten3roberts: It could, but will be remade completely using a schema and exposed to the guest code

crates/network/src/native/client.rs
ten3roberts: Not sure. The `client_state` is shared quite a lot and a bit all over the place.

Haven't looked into how the wasm scripting uses it, but I suspect it may lock it as well

ten3roberts: Will look at it 👍

crates/network/src/web/client.rs
ten3roberts: Yay, more reasons to add logs.

In all seriousness, I will do that. Makes debugging easier for the web now to test and find things that are running before they should (like the renderer panicking if there is no established connection which I've encountered)

crates/std/src/uncategorized/asset_url.rs
ten3roberts: Nope, just missed it trying to figure out how it resolved an asset url to the server address

app/src/client/mod.rs
```diff
@@ -133,11 +133,16 @@ fn MainApp(
...
             server_addr,
             user_id,
-            on_loaded: cb(move |client| {
-                let mut game_state = client.game_state.lock();
+            // NOTE: client.game_state is **locked** and accesible through game_state.
+            //
+            // This is to prevent another thread from updating using the client after connection but
+            // just before `on_loaded`. This is a very small window of time, but does occasionally
+            // happen, especially when joining a server which is already running and server
+            // immediately.
```
pierd, line 144:
```suggestion
            // happen, especially when immediately joining a server which has just been started.
```

ten3roberts: Good catch, though the opposite of what I meant. 

It occurs then the server has been running for a while so there is no loading going on and its server immediately

philpax (approved):

Issue #559, opened 2023-07-04
`git_version` output is not consistent between platforms


philpax:
`git_version` is (indirectly) used by the network protocol to check if the client is on the same version of Ambient as the server. Unfortunately, this is not necessarily consistent between platforms:

> Server version: "0.3.0-dev-c2ed491", Client version "0.3.0-dev-c2ed4913"

We can fix this by

- checking for the common prefix
- finding a version of `git_version!()` that will return the full hash
- using some other kind of versioning identifier

philpax:
Fixed by #566

PR #566, opened 2023-07-06
Fix version formatting in network crate

pierd:
https://github.com/AmbientRun/Ambient/issues/559

`--abbrev=10` - use at least 10 characters from commit hash (should we use the full 40?)
`--always` - always show something, that is at least a commit hash (so it works for commits that don't have any tags in their path to the root)
`--long` - always put all fields, otherwise it will skip the commit hash if the commit is tagged

### DIFF
--- a/app/src/cli/new_project.rs
+++ b/app/src/cli/new_project.rs
@@ -71,7 +71,7 @@ pub(crate) fn new_project(
                 {
                     if let Some(api_path) = api_path {
                         format!("ambient_api = {{ path = {:?} }}", api_path)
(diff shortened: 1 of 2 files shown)


ten3roberts (approved):

ten3roberts:
Great. 

Wasn't aware of that; I just lifted the code from the `new_project_template` module.

Should we change it there as well?

philpax:
Great fix, thanks! Maybe it's worth defining a shared function for this (maybe `ambient_std` or something) so that you can't accidentally use `git_version` without the args?

PR #568, opened 2023-07-07
add hyperlink in ui and text example

chaosprint:


crates/wasm/Cargo.toml
```diff
@@ -20,11 +20,15 @@ ambient_gpu = { path = "../gpu" }
...
 ambient_procedurals = { path = "../procedurals" }
 
+ambient_guest_bridge = { path = "../../shared_crates/guest_bridge", default-features = false, version = "0.3.0-dev" }
+
 ambient_project = { path = "../../shared_crates/project" }
 ambient_shared_types = { path = "../../shared_crates/shared_types", features = [
     "native",
 ] }
 
+webbrowser = "0.8.10"
```
philpax, line 30: Might be a good idea to use `open`, which we have a workspace dep for. Not sure if it'd be a security risk, though - does `open::that("explorer.exe")` work?

shared_crates/ui/src/text.rs
```diff
@@ -68,3 +71,19 @@ pub fn FontAwesomeIcon(
...
+#[element_component]
+/// Hyperlink to a URL. This will open browser when clicked.
+pub fn Hyperlink(
+    _hooks: &mut Hooks,
+    /// The URL to open
+    url: String,
+) -> Element {
+    crate::prelude::Button::new(&url.clone(), move |world| {
+        println!("Opening url {}", &url);
+        let link = Entity::new().with(hyperlink(), url.clone());
```
philpax, line 84: Maybe worth adding a function for this instead of an entity. That would also allow us to return an error if we find the thing that's being opened to be invalid/questionable.

chaosprint: > Maybe worth adding a function for this instead of an entity. That would also allow us to return an error if we find the thing that's being opened to be invalid/questionable.

I tried before to call `webbrowser` here but get a '__wbindgen_placeholder__' error. That's why I have to send to the world and watch it. Perhaps we can do the check there and then decide whether we want to open it or not?

crates/wasm/Cargo.toml
chaosprint: > Might be a good idea to use `open`, which we have a workspace dep for. Not sure if it'd be a security risk, though - does `open::that("explorer.exe")` work?

`open` works.
I am not on Windows so I am not sure of the latter. Isn't `that` only used for URL? https://docs.rs/open/latest/open/

PR #578, opened 2023-07-15
chore: update golden images

philpax:
PR because I'm not sure about some of these. Might need to increase the timeout before screenshot.


ten3roberts:
Why did they need updating?

There are some slight differences, but not really a change made to our source code to warrant updating an example.

Is it just git's pixel perfect hash that deems them different?

philpax:
The biggest issue was with the `third_person_camera` example, which relied on the following chain of events:

- client connects
- client immediately input grabs
- client starts sending input messages to the server
- server receives input messages and adds (not sets) player input components
- server system then kicks in and updates the position of the player, in turn affecting their camera

My recent #575 removed the second step (input grab is no longer automatic), which meant that the messages were never sent and the components were never added, resulting in the test having the wrong image. The second commit fixes that by adding the components ahead of time, so that the system always runs.

The only other major change is the addition of text to the first person camera example. The rest is likely tiny differences in encoding.

ten3roberts:
Fantastic. Always good to know why a change was made, as changes that "where just made because they were made" are often an underlying error.

I've also gotten those other images dirty when doing `golden-images update`.

PR #583, opened 2023-07-17
feat(ui): move more to HooksExt

philpax:
Just letting the CI check I didn't break anything here

### DIFF
--- a/crates/editor/src/ui/mod.rs
+++ b/crates/editor/src/ui/mod.rs
@@ -31,9 +31,9 @@ use ambient_terrain::{
 use ambient_ui_native::{
     command_modifier, height,
     layout::{docking, space_between_items, width, Borders, Docking},
-    margin, use_window_logical_resolution, Button, FlowColumn, FlowRow, FontAwesomeIcon, Hotkey,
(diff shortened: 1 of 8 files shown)


FredrikNoren:
Hm.. this might not be the right direction; the idea with hooks is that they are modular/standalone and that you can write your own, but when going in this direction it might give people the idea that the ones that are defined are the only ones that can exist. Both react and dioxus have free functions for all hooks.

ten3roberts:
I've also found myself in situations where a hook would be perfect for my given situation but I miss it because it was a free function in another crate.

I think that using `Ext` traits for this would be perfect, as rust-analyzer still gives completion for `hooks.us...` even if the Ext trait is not in scope

FredrikNoren:
With dioxus I start typing `use_` and it gives me autocompletion on that:

Screenshot 2023-07-20 at 15 37 26

But yeah, agree that it should be either or, the mix we had now wasn't great. I _think_ it might be the better direction since people won't need to create an Ext trait each time they want to add a hook (though both directions have drawbacks)

philpax:
Yeah I'd be fine with that if anyone wants to pick it up. We might also want to rename `Hooks` to `Context` while we're at it.

PR #589, opened 2023-07-19
feat: concurrent stream handling

ten3roberts:
This will handle the streams on separate tokio/web-sys tasks which prevents new incoming streams from being starved by an in flight stream that is being read.

**Note**: this may induce issues related to assumptions that streams complete in order and that multiple streams are not processed at the same time.

Many of these could be due to short-duration mutex locking and unlocking which may cause logic interleaving in the now concurrent handlers


pierd (approved):
looks good to me :shipit:

Issue #570, opened 2023-07-10
[campfire] commands to build and launch the web client


ten3roberts:
Add commands to campfire to build, serve and launch the web client in its entirely by running all the necessary commands in the correct directory

```
cargo cf example serve [...]
cargo cf web serve-dev
cargo cf web launch-chrome [ip]
```

#151

philpax:
Can we close this issue now that #595 has landed?

ten3roberts:
No. I only had time to add `build` and `open-browser` as I had to move on the CI deploy.

The `serve` is still missing, as well as `open-browser` on linux and windows

philpax:
Just to confirm, it's just windows serve missing now, right?

ten3roberts:
Yes

PR #595, opened 2023-07-21
Campfire web client support

ten3roberts:


ten3roberts:
Fixes: #570

ten3roberts:
I am not going to fully flesh out all the web commands into campfire due to a lack of time as I also need to get the multi-instance-ambient-js-version-manager up and running using this to deploy to an artefact.

The current commands involves building and packaging the application with all relevant shims and configuration, as well as launching chrome with the correct flags to trust our self signed certificate for development

philpax (approved):
Looks good; haven't tested it but I trust that it works. Nice work!

Issue #597, opened 2023-07-22
Make game play window fullscreen, hide all the performance (fps, ms rtt, etc) by default.


chaosprint:
Only show info and make it windowed with some flags

suggested by @TobiasEnholmX

FredrikNoren:
We should _not_ hide the performance metrics by default; this is the development build so I'd expect to see those.

For fullscreen; we can't use esc to exit it, people would likely use that to open the main menu. And if there's a button to exit it, there should also be a button to re-enter it.

Also whatever setting you switch to should be saved, so if I switch to windowed, it should keep that next time I launch ambient. I suspect a lot of people will still want to develop in windowed so just switching to fullscreen for everyone doesn't seem like a good idea.

PR #598, opened 2023-07-22
support fullscreen by default in different os with esc to exit

chaosprint:


chaosprint:
to solve https://github.com/AmbientRun/Ambient/issues/597

ten3roberts:
Here: https://github.com/AmbientRun/Ambient/blob/main/crates/gpu/src/settings.rs

Takes care of the window size, so a fullscreen option could easily be added to this already existing structure to keep consistent

chaosprint:
> Here: https://github.com/AmbientRun/Ambient/blob/main/crates/gpu/src/settings.rs
> 
> Takes care of the window size, so a fullscreen option could easily be added to this already existing structure to keep consistent

It's for resolution, right? I do need to change that one as well. Thanks for the info! The change so far is to make the window "Full Screen" with all the control buttons hidden.

ten3roberts:
Yes, and that would need to be changed since fullscreen (usually) means changing the resolution to the monitor.

I'd suggest making the `resolution` field an `Option<(u32, u32)>` or similar, where `None` is set when going fullscreen meaning "use the monitors resolution"

PR #613, opened 2023-07-25
Server profiling

pierd:
Move `puffin` initialisation so that ambient server can be profiled with it too


ten3roberts (changes requested):
Rejecting this for now, though the changes look good.

The reason being that this will prevent the rust (not guest) examples to be profiled, or other entry points other than `app/main.rs`. I recall that there was a reason we did that in the app rather than main, though not entirely sure what besides the examples, and if that is still relevant to keep.

Just bringing this forth, but if you consider it a worthy tradeoff; go ahead 😄

philpax:
I believe it's so that things that could have performance implications (like the renderer) could be profiled in isolation.

I'd suggest adding an option to the `App` that's a `puffin_port: Option<u16>`, so that it can be specified by the using code (i.e. the Ambient client/server) with its own settings scheme. We could then have `PUFFIN_CLIENT_PORT` and `PUFFIN_SERVER_PORT` env variables that Ambient takes and passes to the `app`.

ten3roberts:
> I believe it's so that things that could have performance implications (like the renderer) could be profiled in isolation.
> 
> I'd suggest adding an option to the `App` that's a `puffin_port: Option<u16>`, so that it can be specified by the using code (i.e. the Ambient client/server) with its own settings scheme. We could then have `PUFFIN_CLIENT_PORT` and `PUFFIN_SERVER_PORT` env variables that Ambient takes and passes to the `app`.

Good idea, having two puffin instances would make the server and client frame interleaving easier, though I am unsure if you can host two puffin instances as the profile spans are a global variable

ten3roberts:
This will enable the physic example to work on the web as well since the web can't depend on the physics crate or its components

pierd:
> Good idea, having two puffin instances would make the server and client frame interleaving easier, though I am unsure if you can host two puffin instances as the profile spans are a global variable

Exactly this, I'm not sure if having 2 puffin instances makes sense here as it's not feasible to separate the spans within the same process. And running separate process (for client and server) already allows for updating env var.

pierd:
That said puffin seems to be frame calculation focused and profiling anything that happens outside of that is a suboptimal experience.

philpax:
Hm, yeah, fair enough. It would be nice to get separate profiling data when using `ambient run`, but I agree that would be pretty difficult to do within the puffin paradigm. This is fine as-is, then!

Issue #606, opened 2023-07-24
`visualize_collider` doesn't visualize


droqen:
Not sure when this started happening.
Repro in newest version of `examples/basics/physics`.

ten3roberts:
Right, tracking it down.

The gizmo renderer is invoked, though there are no gizmos primitives to draw

4c17805 works

PR #616, opened 2023-07-25
Local gizmos

ten3roberts:
Fixes: #606


ten3roberts:
Furthermore, this makes visual debugging easier as gizmos can now be placed directly on an entity. This will in the future allow for e.g; debugging a player or enemy view direction with a line, or raycasts for AFPS

ten3roberts:
This will enable the physics example to work on the web as well since the web can't depend on the physics crate or its components

ten3roberts:
image

crates/std/src/line_hash.rs
```diff
@@ -2,23 +2,10 @@
...
     () => {{
-        let id = concat!(file!(), line!(), column!());
-        line_hash!(id)
-    }};
-    ($($s:expr),*) => {{
-        let mut s: u128 = 0;
-        $(s += $crate::line_hash!($s) as u128;)*
-        $crate::line_hash!(s)
-    }};
+        concat!(file!(), line!(), column!())
```
philpax, line 9: Hm, this isn't a hash; I don't know if that actually matters but maybe we should rename it to `line_uid` or something

crates/rect/src/lib.rs
```diff
@@ -85,6 +89,24 @@ pub fn systems() -> SystemGroup {
...
         vec![
+            ensure_has_component_with_default(pixel_line_from(), line_from()),
+            ensure_has_component_with_default(pixel_line_to(), line_to()),
+            query_mut(
+                (line_from(), line_to()),
+                (pixel_line_from().changed(), pixel_line_to().changed()),
+            )
+            .to_system(|q, world, qs, _| {
+                if let Some(world_cam) = Camera::get_active(world, main_scene(), None) {
+                    if let Some(ui_cam) = Camera::get_active(world, ui_scene(), None) {
```
philpax, line 100: Could potentially simplify this as either
`(Some(world_cam), Some(ui_cam)) = (..., ...)`
or
`Some((world_cam, ui_cam)) = (...).zip(...)`

shared_crates/schema/src/schema/physics.toml
```diff
@@ -157,5 +157,7 @@ attributes = ["Debuggable", "Networked", "Store"]
 [components."core::physics::visualize_collider"]
 type = "Empty"
 name = "Visualize collider"
-description = "If attached, the collider will be rendered."
+description = """If attached, the collider will be rendered.
+**Note**: this will hijack `local_gizmos`
```
philpax, line 161: Clarify what this means - does it replace it entirely or add to it?

philpax (changes requested):
Looks pretty reasonable to me. I've made a few minor suggestions around code style, but implement as you please.

The gizmos can't currently be used by the guest code, right?

crates/rect/src/lib.rs
ten3roberts: Ok, just lifted this code from the physics visualization

shared_crates/schema/src/schema/physics.toml
ten3roberts: By hijack I mean a system which consistently overwrites the component, so if you store your own things there it will be lost all the time. Think writing to position while you have velocity, etc

crates/std/src/line_hash.rs
ten3roberts: Good point. Realized it would be far more descriptive to find out the origin of a gizmo scope rather than prehash it

shared_crates/schema/src/schema/physics.toml
philpax: Mention that in the component description and we're good to go

philpax (approved):
lgtm, just mention the hijack behaviour in more detail in the comp description

PR #619, opened 2023-07-25
feat: optimize files generated by audio build pipeline with OptiVorbis

AlexTMjugador:
[OptiVorbis](https://github.com/OptiVorbis/OptiVorbis) is a novel Ogg Vorbis file optimization tool that slightly reduces their size without any compromise on audio quality, portability, or runtime decoding performance. It executes much faster than most generic compression programs, so it shouldn't affect build times too much.

OptiVorbis has been thoroughly tested, and its documentation explains in detail how it works. Given that audio assets can take a substantial chunk of the total game size, I think OptiVorbis is a great fit for Ambient and has the potential to substantially improve the distribution experience for players.

Disclaimer: I am the author of OptiVorbis. Thanks for using the `vorbis_rs` crate, which I also wrote, and which made me aware of this cool game engine project!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,6 +90,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
+
+[[package]]
 name = "alsa"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -394,8 +400,8 @@ dependencies = [
  "image",
  "itertools",
  "log",
+ "optivorbis",
  "parking_lot",
- "rand 0.8.5",
  "relative-path",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


chaosprint:
Thanks for the PR. LGTM and tested locally.

Pombal:
@AlexTMjugador I've noticed that optivorbis has git2 as a build dependency which builds libgit2-sys. libgit2-sys takes 2 minutes to build on a modern windows machine. Would it be possible to remove this dependency from optivorbis or to make it optional?

AlexTMjugador:
Yes, it is possible. The only reason it has that dependency is to automatically retrieve and embed nice version information in its CLI. Refactoring that dependency is on my to-do list, but I don't know when I'll be able to get to it.

Pombal:
I appreciate the quick reply!
I've reverted the change in https://github.com/AmbientRun/Ambient/pull/751 because we're making a concerted effort to improve developer iteration times but we'll reintroduce the change once compile times improve.

AlexTMjugador:
I have just removed the build-time dependency on `git2` and `time` in this commit: https://github.com/OptiVorbis/OptiVorbis/commit/a7e7983e7bd65108da3bcfdbf7b54805a58e754e. If it's okay with you, I can submit a PR to bring OptiVorbis back into Ambient once this change makes it into a release. In the meantime, feel free to check it out and give feedback :+1:

Pombal:
That's perfect, thanks! A PR to bring it back would be much appreciated :)

PR #621, opened 2023-07-26
Deploy ambient web client

ten3roberts:


.github/workflows/ci.yml
```diff
@@ -103,6 +103,34 @@ jobs:
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --package campfire
+
+      - name: "Build web client"
+        uses: actions-rs/cargo@v1
+        with:
+          command: run
+          args: --package campfire -- web build --profile release --target standalone
```
ten3roberts, line 126: The `--target standalone` is not necessary.

Removing it will produce a node package (just a package.json added), I kept this here to be consistent and do the exact same parameters as the deploy workflow

campfire/src/web/build.rs
```diff
@@ -19,31 +14,75 @@ pub(crate) enum Target {
...
+    }
+
+    Ok(())
+}
+
+pub fn ensure_wasm_pack() -> anyhow::Result<()> {
     match which::which("wasm-pack") {
         Err(_) => {
-            eprintln!("Installing wasm-pack");
-            let status = Command::new("carg")
```
pierd, line 39: Was there a never spotted typo here?

pierd (approved):
YAML LGTM

campfire/src/web/build.rs
ten3roberts: Good catch. It wasn't spotted as that is the macos install

PR #622, opened 2023-07-26
Despawn player recursively

pierd:


philpax (approved):
I think this should work fine, but let's wait a bit before merging it just to confirm there aren't any scenarios in which recursively despawning the player would have negative consequences

ten3roberts:
> I think this should work fine, but let's wait a bit before merging it just to confirm there aren't any scenarios in which recursively despawning the player would have negative consequences

What comes to my mind is animations, both the player, but also other systems, embers or callbacks which execute just after the same frame, depending on registration order

ten3roberts (approved):

PR #625, opened 2023-07-28
Allow web client to join a custom url

ten3roberts:


philpax (approved):
Seems pretty reasonable, but I'd suggest `campfire-no-openssl` over `campfire-ndf` to make it more obvious why that's being used, and where that can't be used (i.e. the web build CI, adding comments that explain why no default features are being used). Otherwise, LGTM!

ten3roberts:
I made it ndf (maybe --no-default) is better so that we can add more feature gates. `--no-ssl` is thus no longer relevant.

For example, building the *doc* features with `rustdoc-types` and `cargo_toml` etc are heavy an absolutely not required for the CI

philpax:
Maybe `slim` then instead? "no default" doesn't really convey that much information about what the conceptual change is

PR #636, opened 2023-08-04
add import for cli assets command

chaosprint:
@ten3roberts I have to change a little bit on your previous implementation on Migrating. I hope it makes sense.

### DIFF
--- a/app/src/main.rs
+++ b/app/src/main.rs
@@ -280,10 +280,10 @@ async fn main() -> anyhow::Result<()> {
                         // TODO: import textures API may change, so this is just a placeholder
                         todo!();
                     } else {
-                        eprintln!("Unsupported file type");
+                        bail!("Unsupported file type");
                     }
                 }
-                None => eprintln!("Unknown file type"),
+                None => bail!("Unknown file type"),
             },
         }
 
--- a/app/src/main.rs
+++ b/app/src/main.rs
@@ -277,6 +277,7 @@ async fn main() -> anyhow::Result<()> {
                         )
                         .context("failed to import models")?;
                     } else if ext == "jpg" || ext == "png" || ext == "gif" || ext == "webp" {
+                        // TODO: import textures API may change, so this is just a placeholder
                         todo!();
                     } else {
                         eprintln!("Unsupported file type");
--- a/app/src/main.rs
+++ b/app/src/main.rs
@@ -254,16 +254,36 @@ async fn main() -> anyhow::Result<()> {
         return Ok(());
     }
 
-    if let Commands::Assets { command, path } = &cli.command {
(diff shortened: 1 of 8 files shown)


app/src/main.rs
```diff
@@ -254,16 +254,37 @@ async fn main() -> anyhow::Result<()> {
...
+                        )
+                        .context("failed to import models")?;
+                    } else if ext == "jpg" || ext == "png" || ext == "gif" || ext == "webp" {
+                        // TODO: import textures API may change, so this is just a placeholder
+                        todo!();
+                    } else {
+                        eprintln!("Unsupported file type");
+                    }
+                }
+                None => eprintln!("Unknown file type"),
```
ten3roberts, line 286: I think it would be better to fail and return an error, rather than just printing text and exiting with SUCCESS.

Something like `=> anyhow::bail!("Unknown file type")` etc.

Silent errors that proceed anyways can be very frustrating since you easily miss a single line in the terminal if the rest of the command proceeds.

crates/build/src/pipelines/importer.rs
```diff
@@ -0,0 +1,141 @@
...
+    let mut file = std::fs::File::open(&tomlpath)?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+
+    let mut data: toml::Value = match toml::from_str(&contents) {
+        Ok(v) => v,
+        Err(_) => toml::Value::Table(toml::map::Map::new()), // if we cannot parse the file, start with a fresh table
+    };
+
+    if let toml::Value::Table(table) = &mut data {
```
ten3roberts, line 27: I think I'd prefer to add `use toml::Value` and `use std::path::Path` etc to reduce the noise in the function body 😊

crates/build/src/pipelines/importer.rs
```diff
@@ -0,0 +1,141 @@
...
+        .open(tomlpath)?;
+    write!(file, "{}", toml::to_string(&data)?)?;
+
+    let file_name = path.file_name().unwrap(); // get the file name from the path
+    let destination = asset_folder_path.join(file_name);
+    std::fs::copy(path.clone(), destination).context("Error copying audio file")?;
+    Ok(())
+}
+
+pub fn import_model(path: PathBuf, collider_from_model: bool) -> anyhow::Result<()> {
```
ten3roberts, line 68: Same in here.

No need to repeat `std` all the time, it makes it more difficult to read

ten3roberts (changes requested):
Looks exciting.

Have some few suggestions that you could address if you like

chaosprint:
> Looks exciting.
> 
> Have some few suggestions that you could address if you like

Done. Thanks!

Issue #628, opened 2023-07-31
`EnitityId` is serialized as a base64 encoded string on the wire


pierd:
Internally it's just `u128` so we are artificially creating 29 bytes out of 16.

ten3roberts:
I do believe it was like this originally to aid debugging when viewing the serialized world as a file or through logs.

It also helps when comparing network traffic with other sources such as the ecs debugger as entity IDs Debug impl uses the same representation.

pierd:
I've been looking into how we are using entity ids. And I'd like to make them smaller or at least smaller on the wire.

From what I've seen we can kind of split entity ids into 3 groups depending were they originate from:
1. Generated during build and stored in asset descriptions (for example in main.json in models and prefabs).
2. Generated on the server.
3. Generated on clients.

Is it correct that entities originating from the client (3.) are never shared with the server or other clients?
Assuming that most of the entities referenced on the wire are going to be the ones generated on the server (2.), I'd like to use sequential ids for them and use [VarInt encoding](https://docs.rs/bincode/1.3.1/bincode/config/struct.VarintEncoding.html). This would shorten most entity ids on the wire from 29 bytes to <5 bytes.

FredrikNoren:
@pierd Sequential won't work, because they need to be creatable distributed (i.e. you can create an id on the client). We had generational before (kind of like sequential), but switched for that reason. Here's the issue where I was working on the switching: https://github.com/AmbientRun/Dims/issues/543

Issue #627, opened 2023-07-31
Improve component path serialisation in diff stream


pierd:
Currently we serialise the whole component path as a string. For example "core::physics::angular_velocity" or "core::transform::translation". So a `WorldDiff` can have say 137 `WorldChange::Set`s with paths being 3290 bytes in total but have only 7 unique paths with total length of 174 bytes (taken from a `WorldDiff` seen on wire).

We could really save a lot of bandwidth here. Also given the fact that a string in bincode is serialised as `u64` length followed by the string bytes (for example "core::physics::angular_velocity" becomes 39 bytes: `\x1f\0\0\0\0\0\0\0core::physics::angular_velocity`).

PR #660, opened 2023-08-09
Implement diff deduplication

pierd:
- filters out `WorldChange::Set` operations if they already existed in the previous diff
- skips sending `WorldChange::Set` operations for removed entities/components (previously we would send them but they would silently fail since they are applied as last ones)
- changes `WorldChange::Set` into `WorldChange::SetComponents` for updating multiple components of the same entity at once
- removes `create_revert` flag from `WorldDiff` and `WorldChange` - dead code and I didn't want to create an extra implementation for `WorldChange::SetComponents`

With these changes `afps` uses ~100kB/s instead of ~4000kB/s when all blocks are destroyed.


crates/network/src/diff_serialization.rs
```diff
@@ -0,0 +1,57 @@
...
+                    // we only populate components_to_drop for SetComponents changes
+                    unreachable!();
+                }
+            }
+            keep
+        });
+        self.last_diff = new_diff;
+        diff
+    }
+}
```
pierd, line 57: New module. I plan to put more serialisation specific code here (for #627 and #628).

FredrikNoren (approved):
Neat! The only thing I want to request is an update to the book with info about this so people know what to expect

docs/src/reference/networking.md
```diff
@@ -16,6 +16,12 @@ The client is fundamentally designed around runtime flexibility of logic, which
 
 If on 0.2 or above, consult the [clientside](https://github.com/AmbientRun/Ambient/blob/main/guest/rust/examples/intermediate/clientside/ambient.toml) example to see how to define networked components.
 
+### Entities synchronization
+
+The Ambient runtime synchronizes entities using a diff-based approach. The server sends a `WorldDiff` to the client, which contains a list of entities to spawn and despawn, and components to add, update, and remove. Note that some operations might be batched for performance or not included in the update sent to the clients if they effectively don't change any values. For example adding 0 to a number or changing a boolean to false and back to true within the same frame might not emit an update and might not trigger a `change_query`. We recommend using messaging if such events are important to your game.
+
+Currently the client applies the changes to its local world as soon as they are received.
+
```
pierd, line 24: @FredrikNoren how does it sound?

crates/ecs/src/stream.rs
```diff
@@ -204,38 +190,37 @@ pub struct FrozenWorldDiff {
...
-                        continue;
+        for change in diffs.iter().flat_map(|diff| diff.changes.iter()) {
+            if let WorldChange::SetComponents(entity_id, entity) = change {
+                if let Some(idx) = set_idx.get(entity_id) {
+                    if let Some(WorldChange::SetComponents(_, existing_entity)) =
+                        changes.get_mut(*idx).map(Cow::to_mut)
+                    {
+                        existing_entity.merge(entity.clone());
+                    } else {
+                        panic!("Expected SetComponents");
```
philpax, line 206: Can you extend this panic/add a comment about the invariants that would prevent this panic from firing? (Or if it *can* fire, can you add more information about what to look out for?)

philpax (approved):

crates/ecs/src/stream.rs
pierd: It should actually be `unreachable!()`. I'll change it and add a comment.

pierd: https://github.com/AmbientRun/Ambient/commit/e3318819325dd26ce1f317b39d4e72723b9ca10e

PR #687, opened 2023-08-14
add anim import helper and avoid repeat import

chaosprint:


app/src/main.rs
```diff
@@ -99,6 +99,12 @@ fn main() -> anyhow::Result<()> {
                         } else {
                             bail!("Unsupported file type");
                         }
+                        ambient_build::build_assets(
+                            &assets,
+                            &PathBuf::from("assets"),
+                            &PathBuf::from("build"),
```
philpax, line 105: Will this work if your project path is elsewhere, or if you have a multi-ember project?

chaosprint: > Will this work if your project path is elsewhere, or if you have a multi-ember project?

Not sure. I will try to get used to the new ember layout ASAP and test this.

chaosprint: > Will this work if your project path is elsewhere, or if you have a multi-ember project?

I just tested it on AFPS and it works but the `.gitignore` template has to be updated:

https://github.com/AmbientRun/Ambient/pull/687/files#diff-b66ef7667ee679cc1feafc8df5009b9075b8c8db41e7aaa437781c6b964526a9

I hope it makes sense as with a new ember system, it's possible that we just copy an ember that contains a `build` there.

PR #697, opened 2023-08-15
fix #690 - remove `Entity::with_default`

### LINKED ISSUE #690: ban `.with_default`
Following discussion about Bool/Empty components here https://github.com/AmbientRun/Ambient/issues/577 I ceased using `.with_default` and have found it clears up a lot of footguns without really slowing down my workflow at all.

Components that use `Quat` and `Mat4` were initially slightly annoying but learning to type `Quat::IDENTITY` helped me to understand the codebase a little more, rather than hiding something so that I never understood it (or potential problems) in the first place. It's not a problem anymore, and the transition was a good one with basically no drawbacks for me.

I recommend not using `.with_default` at all, and changing at least the auto-generated `server.rs` code to not use it either.

philpax:
After discussion with others, and from my own personal experience, I decided that we should remove `with_default` because its name suggests something it's not (i.e. it's the default value for that component, not the default value for the type).

I've also removed it from the host API to ensure that it doesn't accidentally bleed through in the shared code between the two.

No deprecation as we've already made severe breaking changes, so we might as well add onto the pile 🤷 

See #690 for more context.


FredrikNoren (approved):
Seems reasonable.

I _suspect_ that we'll get feature requests for this in the future though. A better implementation might be one where only some specific components have defaults that are actually values specified in the ambient.toml, instead of all of them having defaults that are rust value defaults.

philpax:
Yeah, that was what I was thinking as well - we already sort of support component defaults, so taking that support all the way would work well. I'll make another issue for it.

philpax:
Golden images pass locally, so I'm going to merge.

PR #699, opened 2023-08-15
Adjust debug/release build defaults

pierd:
This adds `--debug` flag (next to the existing `--release`) so that commands can have different defaults (while still allowing for an override):
- `run` - debug (good for local development)
- `build` - debug
- `serve` - release (remote players might be joining - smaller file sizes preferred)
- `deploy` - release (IMHO the most sane default for this case)

The idea is that release builds give us way smaller WASM files so that makes them a good default for whenever those files are going to be uploaded (either via deploying, proxying or serving).


app/src/cli/mod.rs
```diff
@@ -148,9 +148,13 @@ pub struct ProjectCli {
     /// The path or URL of the project to run; if not specified, this will default to the current directory
     pub path: Option<String>,
 
-    /// Build all the assets with full optimization; this will make debugging more difficult
+    /// Build all the assets with debug information; this will make them larger (default for all commands apart from `deploy` and `serve`)
+    #[arg(short, long, conflicts_with = "release")]
+    debug: bool,
+
+    /// Build all the assets with full optimization; this will make debugging more difficult (default for `deploy` and `serve`)
```
philpax, line 155: I'd suggest mentioning both the pros and cons for both of these comments (i.e. debug is "larger and slower but easier to debug" and release is "smaller and faster but harder to debug"), just so that you don't have to look at the other option to see the downside

app/src/cli/mod.rs
```diff
@@ -242,4 +246,30 @@ impl Cli {
...
         }
     }
+    pub fn use_release_build(&self) -> bool {
+        match &self.command {
+            Commands::New { .. } => false,
+            Commands::Run { project_args, .. } => project_args.is_release().unwrap_or(false),
+            Commands::Build { project_args, .. } => project_args.is_release().unwrap_or(false),
+            Commands::Deploy { project_args, .. } => project_args.is_release().unwrap_or(true),
+            Commands::Serve { project_args, .. } => project_args.is_release().unwrap_or(true),
+            Commands::View { project_args, .. } => project_args.is_release().unwrap_or(false),
```
philpax, line 256: You might be able to do `Commands::Run { project_args, .. } | Commands::Build { project_args, .. }` etc - as long as the bound variables are the same, you can group them together.

philpax (approved):
lgtm, made a few minor comments but nothing blocking

PR #702, opened 2023-08-15
Reduces the verbosity of campfire command failures

ten3roberts:


crates/wasm/src/client/mod.rs
```diff
@@ -9,6 +9,7 @@ pub fn initialize(
     world: &mut World,
     messenger: Arc<dyn Fn(&World, EntityId, shared::MessageType, &str) + Send + Sync>,
 ) -> anyhow::Result<()> {
+    // anyhow::bail!("test");
```
philpax, line 12: prolly don't need this?

philpax (approved):

ten3roberts:
Right, fixed.

Snuck in another change which removes the `.env` as having it auto-populate environment variables is unexpected and may lead to some inconsistent results (the classic statement of "explicit rather than implicit). I was the only one that used the file as well, as I don't think vscode auto loads `.env` files

Issue #692, opened 2023-08-15
Failed to join session with deserialization error


Pombal:
ERROR ambient_network::native::client] Game failed: Failed to deserialize payload of type ambient_network::proto::ServerPush

    Caused by:
        string is not valid utf8: invalid utf-8 sequence of 1 bytes from index 172

ten3roberts:
Which game did you play? Was it over a proxied or hosted connection?

pierd:
Do you know the builds of ambient server and client?

Pombal:
Latest main as of yesterday at around 6pm

pierd:
Ok, I've seen it happening. I'm going to double check but I think `ServerPush` schema changed internally (somewhere inside `ExternalComponentDesc`) and now we fail to deserialize it if different versions are used on client and server. And since the server version is passed in that message, we can't even print the usual warning.

pierd:
I'll make sure that we can at least deserialise part of this struct to at least have some more meaningful message instead of that cryptic one.

PR #706, opened 2023-08-16
Handle external components deserialization issues gracefully

pierd:
For situations like this: https://github.com/AmbientRun/Ambient/issues/692


ten3roberts (changes requested):

pierd:
@ten3roberts thanks, newtyped Result makes much more sense. Updated, could you take another look?

philpax (approved):
LGTM. Maybe add a note that `FailableDeserialization` has to be the last field in your deserialization _or_ that anything after it will likely have failed deserialization as well?

Issue #684, opened 2023-08-14
Ambient version manager cli


FredrikNoren:
- We rename `ambient` to `ambient_runtime`
- We create a new crate called `ambient`
- To install; `cargo install ambient`

`ambient new hello` // this defaults to latest version

In `ambient.toml`; we add `ambient_version = 0.2.1`

`ambient run`
1. Check the version in ambient.toml -> 0.2.1.
2. Check if we have it downloaded, if not, we download that ambient_runtime version and put it in some cache folder (https://crates.io/crates/directories)
3. Copy arguments to `ambient_runtime` from `ambient`. So for instance if we use `ambient run --headless` -> `[cache_folder]/ambient_runtime_0_2_1 run --headless`

## For nightly builds

Create .github/nightly.ci which builds the runtime each night, and then deploys it to a storage bucket.

`ambient --version=nightly-2023-08-14 new` // download from the nightly storage bucket

FredrikNoren:
Done

PR #708, opened 2023-08-16
Add `/info` endpoint

pierd:
for https://github.com/AmbientRun/Ambient/issues/684


philpax (approved):
it might be nice to have a helper somewhere for `ambient_version() -> AmbientVersion { version: String, revision: String }` so that we're using the same logic everywhere, but lgtm, not a blocker

PR #712, opened 2023-08-17
show rev in cli version

chaosprint:
use cli like before but add a rev in version display; easier to know the exact version

image

### DIFF
--- a/app/src/cli/mod.rs
+++ b/app/src/cli/mod.rs
@@ -14,9 +14,9 @@ mod project_path;
 pub use project_path::*;
 
 use self::assets::AssetCommand;
-use ambient_native_std::{ambient_version, git_revision};
+use ambient_native_std::ambient_version;
 #[derive(Parser, Clone)]
-#[command(author, version = {format!("{} (rev {})", ambient_version().version, git_revision().unwrap_or("unknown".to_string()))}, about, long_about = None)]
+#[command(author, version = { let v = ambient_version(); format!("{} (rev {})", v.version, v.revision) }, about, long_about = None)]
 #[command(propagate_version = true)]
 pub struct Cli {
     #[command(subcommand)]
--- a/app/src/cli/mod.rs
+++ b/app/src/cli/mod.rs
@@ -14,9 +14,9 @@ mod project_path;
 pub use project_path::*;
 
 use self::assets::AssetCommand;
-
+use ambient_native_std::{ambient_version, git_revision};
 #[derive(Parser, Clone)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version = {format!("{} (rev {})", ambient_version().version, git_revision().unwrap_or("unknown".to_string()))}, about, long_about = None)]
 #[command(propagate_version = true)]
 pub struct Cli {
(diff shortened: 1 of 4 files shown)


app/build.rs
ten3roberts, line 1: Don't we already have a macro which returns the git commit hash? Except it also makes sure it is consistent?

app/src/cli/mod.rs
```diff
@@ -16,11 +16,13 @@ pub use project_path::*;
 use self::assets::AssetCommand;
 
 #[derive(Parser, Clone)]
-#[command(author, version, about, long_about = None)]
```
philpax, line 19: You can specify an expression to use for `version` here instead of adding `--version` as a bool: https://docs.rs/clap/latest/clap/_derive/index.html#command-attributes

philpax (changes requested):

app/src/cli/mod.rs
chaosprint: thanks! what a detour I made 😅

app/src/cli/mod.rs
```diff
@@ -14,9 +14,9 @@ mod project_path;
 pub use project_path::*;
 
 use self::assets::AssetCommand;
-
+use ambient_native_std::{ambient_version, git_revision};
 #[derive(Parser, Clone)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version = {format!("{} (rev {})", ambient_version().version, git_revision().unwrap_or("unknown".to_string()))}, about, long_about = None)]
```
philpax, line 19: Not sure if you can do arbitrary expressions here, but it might be worth trying to do
`version = { let v = ambient_version(); format!("{} (rev {})", v.version, v.revision) }`

crates/native_std/src/lib.rs
```diff
@@ -163,13 +163,15 @@ pub fn git_revision_full() -> Option<String> {
...
     pub version: String,
     pub revision: String,
+    pub revision_full: String,
 }
 
 impl Default for AmbientVersion {
     fn default() -> Self {
         Self {
             version: env!("CARGO_PKG_VERSION").to_string(),
-            revision: git_revision_full().unwrap_or_default(),
```
philpax, line 172: Sorry to pessimize, but is there anything that relied on it being the full string before this?

ten3roberts: Adding to this, what are the cases in which it can panic? Having the revision silently become `unknown` will just fail in even weirded ways further down the line, as it will try to use a branch called `unknown` instead of a revision etc

ten3roberts (changes requested):

crates/native_std/src/lib.rs
ten3roberts: > Sorry to pessimize, but is there anything that relied on it being the full string before this?

Wasn't this because of the git commit hash being cut short if it was proven by git to be unique based on the other commits in the tree?

It varied based on how many branches were pulled, so we got version mismatches on the commit hash being different lenghts depending on which commits were proven to be unique enough based on which commits were in the tree on the build VM

PR #718, opened 2023-08-17
Optimize serialized diff size

pierd:
- indexes paths and only sends u32 index (plus newly encountered paths) - https://github.com/AmbientRun/Ambient/issues/627
- serializes EntityId in binary format - https://github.com/AmbientRun/Ambient/issues/628
- switches to var int encoding to save space on encoded lengths and other small integers

### Measured by comparing old serialized size to the new one (this PR):

Longer test with afps:
```
old:
    1KB:   976
    2KB:  1156
    4KB:   385
    8KB:  1164
   16KB:  1783
   32KB:   680
   64KB:    36
Average: 10KB

new:
    1KB:  2352
    2KB:  1064
    4KB:  1610
    8KB:  1126
   16KB:    28
Average: 3KB
```

afps after destroying boxes and rejoining to see difference on initial world diff:
```
old:
    1KB:   259
    2KB:   742
    4KB:    85
    8KB:    79
   16KB:   132
   32KB:    79
   64KB:     3
  128KB:     0
  256KB:     0
  512KB:     0
 1024KB:     1
Average: 6KB

new:
    1KB:  1061
    2KB:    82
    4KB:   134
    8KB:   100
   16KB:     2
   32KB:     0
   64KB:     0
  128KB:     1
Average: 1KB
```

### DIFF
--- a/crates/network/src/diff_serialization.rs
+++ b/crates/network/src/diff_serialization.rs
@@ -1,7 +1,22 @@
 use std::collections::HashMap;
 
-use ambient_ecs::{EntityId, WorldChange, WorldDiffView};
+use ambient_ecs::{
+    with_component_registry, ComponentDesc, ComponentEntry, Entity, EntityId, Serializable,
+    WorldChange, WorldDiff, WorldDiffView,
+};
+use bincode::Options;
 use bytes::Bytes;
+use serde::{
+    ser::{SerializeSeq, SerializeTuple},
+    Deserialize, Deserializer,
+};
+
+/// Bincode options used for diff serialization
+pub fn bincode_options() -> impl Options {
+    bincode::DefaultOptions::new()
+        .with_varint_encoding()
+        .allow_trailing_bytes()
+}
 
 #[derive(Clone, Debug, Default)]
 pub struct WorldDiffDeduplicator {
@@ -19,7 +34,13 @@ impl WorldDiffDeduplicator {
                     for entry in entity.iter() {
                         let key = (*id, entry.desc().index());
                         // currently comparing serialized bytes since we don't have cmp for components, could be improved
-                        let bytes: Bytes = bincode::serialize(entry).unwrap().into();
+                        let ser = entry
+                            .attribute::<Serializable>()
+                            .expect("diff should only have serializable components");
+                        let bytes: Bytes = bincode_options()
+                            .serialize(ser.serialize(entry))
+                            .unwrap()
+                            .into();
                         new_diff.insert(key, bytes.clone());
                         if self.last_diff.get(&key) == Some(&bytes) {
                             duplicates.push(entry.desc());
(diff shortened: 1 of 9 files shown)


crates/network/src/diff_serialization.rs
```diff
@@ -54,3 +75,675 @@ impl WorldDiffDeduplicator {
...
+            NetworkedWorldChange::Spawn(_, _) => WorldChangeTag::Spawn,
+            NetworkedWorldChange::Despawn(_) => WorldChangeTag::Despawn,
+            NetworkedWorldChange::AddComponents(_, _) => WorldChangeTag::AddComponents,
+            NetworkedWorldChange::RemoveComponents(_, _) => WorldChangeTag::RemoveComponents,
+            NetworkedWorldChange::SetComponents(_, _) => WorldChangeTag::SetComponents,
+        }
+    }
+}
+
+impl TryFrom<u8> for WorldChangeTag {
```
ten3roberts, line 114: Minor nitpick.


This could be placed immediately after the enum to make it harder to miss to update if you touch the enum. It's current far away state may cause it to be missed when glancing through the code

crates/network/src/diff_serialization.rs
```diff
@@ -54,3 +75,675 @@ impl WorldDiffDeduplicator {
...
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        formatter.write_str("enum WorldChange")
+    }
+
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: serde::de::SeqAccess<'de>,
+    {
+        // WorldChange is encoded as a tuple of 3 elements:
```
ten3roberts, line 285: :heart: love comments about "implicit" protocols

crates/network/src/diff_serialization.rs
```diff
@@ -54,3 +75,675 @@ impl WorldDiffDeduplicator {
...
+    fn visit_u32<E>(self, v: u32) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        match self.known_component_paths.get(&v) {
+            Some(path) => {
+                let component = with_component_registry(|r| r.get_by_path(path));
+                match component {
+                    Some(desc) => Ok(desc),
+                    None => Err(serde::de::Error::custom(format!("No such component: {v}"))),
```
ten3roberts, line 493: Is there any sane path in which this can be triggered.

My only idea is if receiving and deserialization occurs before the *ServerInfo* recv future has been completed

crates/network/src/proto/server.rs
```diff
@@ -372,21 +375,26 @@ pub async fn handle_diffs<S>(
...
                 diffs_count = diffs.len(),
                 merged_diffs_count,
                 final_diffs_count = final_diff.changes.len(),
                 bytes = msg.len(),
             );
-            debug_assert!(
-                bincode::deserialize::<WorldDiff>(msg.as_ref()).is_ok(),
-                "Merged diff should deserialize as WorldDiff correctly"
-            );
+            #[cfg(debug_assertions)]
```
ten3roberts, line 389: I assume this is sort of an inlined test in debug builds, that tests that roundtrips work for each outgoing diff?

ten3roberts (changes requested):
Looks absolutely amazing.

I especially like the use of traits and functional mappings to make the relationship and conversions between the different diff types manageable.

Left some small questions and comments, otherwise all is good.

crates/network/src/proto/server.rs
pierd: It's equivalent to the previous debug_assert, the idea is to fail fast but since it's not a cheap operation then we do it only in debug builds.

I'm a big fan of asserts and failing fast.

FredrikNoren (approved):
Awesome!

crates/network/src/diff_serialization.rs
pierd: I mostly followed the pattern of the regular serialization implementation.

I'd say it can either be caused by broken deserialization (for example subtle error when mismatched versions are used) or some timing issue as you mentioned.

PR #734, opened 2023-08-23
chore: update rustls to fix dependabot alerts

philpax:
Quick test seems to indicate this works fine, but making it a PR just so that @ten3roberts can verify.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1108,7 +1108,7 @@ dependencies = [
  "profiling",
  "quinn",
  "rand 0.8.5",
- "rustls 0.21.1",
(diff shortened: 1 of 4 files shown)


ten3roberts:
Fantastic

https://github.com/rustls/rustls/commit/697846460dd1652529ec33a098e43b88ac3a85c8

The rename is only to clarify the general purpose of the trust anchors, as they aren't server specific. It is not related to any CVE or other fix that we have to keep in mind

ten3roberts (approved):
The good ol' saying LGTM, merge if and when CI passes

PR #742, opened 2023-08-23
Run CI build job in debug mode

Pombal:
Test how much it improves CI build times


Pombal:
Saves 7 minutes build time on windows and mac and loses 4 minutes on golden image tests.
I think this is a worthwhile tradeoff given that we have more control over optimizing the performance of the tests.

PR #747, opened 2023-08-25
fix(web): semantic packages

ten3roberts:
Fixes `Error: no such component: 336` seen on the web client


ten3roberts:
image

philpax (approved):
It would be ideal to share the component inits between the clients, but this is fine for now

ten3roberts:
> It would be ideal to share the component inits between the clients, but this is fine for now

Yes. I am working on that now, as all the component inits uses libraries which are either currently not supported on wasm, or never will be, such as physics

ten3roberts:
Golden images and the failing examples all work locally (Mac Mini S2)

PR #661, opened 2023-08-09
Clientside scripting on the web

ten3roberts:

PR #752, opened 2023-08-25
Removing wasmtime's cache feature also removes the dependency on zstd which has a significant effect on compile times.

Pombal:


ten3roberts:
This will likely conflict with #661 as it changes the wasmtime dependency altogether and no longer uses it

philpax (approved):
TIL about `cache`. It would be nice to get that working at some point to avoid the JIT compilation hit if you're launching the same modules again, but it seems underdocumented so I'm guessing it's not really there yet. LGTM otherwise.

Pombal:
@ten3roberts Are you ok with this change landing and resolving the conflict once client-wasm in merged?

ten3roberts:
Yes, I can resolve the merge conflict absollutely. It'll just unfortunately noop when my branch is merged and we will have to add it to `wasm-bridge` instead, but I'll migrate the side effects of this change over 😊

PR #753, opened 2023-08-25
Remove debug info to improve iteration times

Pombal:
The reasoning is that debug info slows down build times for everyone but isn't heavily used. It can be enabled on a need to basis. But if this will severaly impact some workflows, I'm happy to not to go ahead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,3 +213,7 @@ wasm-bindgen-shared = { git = "https://github.com/rustwasm/wasm-bindgen.git" }
 
 [profile.release]
 incremental = true
+
+[profile.dev]
+# Debug info slows down compile times and isn't heavily used
+debug = 0


ten3roberts:
I know most of us mac dwellers run ambient mostly through dap, and as such utilize debug info for catching and stopping on panics and breakpoints. I mostly use a normal command invocation, but I know both Fredrik and Mithun use the `F5` key with vscode to launch with a debugger present to be able to stop directly at a panic

Additionally, rust panic messages will get context lost or incorrect if debug info is disabled IIRC

When we add sentry integration (which from what I've understood is something we are looking into for the launch) we would require debug info for usable alerts.

Perhaps there is a way we can add another cargo profile for `dist` targets to enable them there, though that would mean we miss out on all the people compiling ambient from source or `cargo install --git https://github.com/AmbientRun/Ambient app`

ten3roberts:
Or perhaps I am misunderstanding what you mean by "not heavily used" :sweat_smile: 

Anyhow, it is great that you are looking into compile times and vendoring our dependencies; it is something I've had on my mind for a long time as we have many different dependencies which achieve the same task

Pombal:
By not heavily used I mean I'm not aware of everyone's workflows :)
But I haven't seen most people use a debugger, at least not as heavily as I'm used to in C++ land. And given that everyone, including Alex, needs to compile code but not make use of debug symbols, I thought that perhaps the tradeoff would make sense. When people need to use a debugger they would debug symbols manually.

As for Sentry, I hope we'll only distribute release builds to users (master actually), not debug builds.

With that said, it's quite likely that this is not a good idea, hence why I'm asking for everyone's opinion.

philpax:
Hm, yeah, not sure about this one - our users (i.e. Alex) use release/production builds (using `cargo install`), so it doesn't impact them too badly. On the other hand, even though I don't use a debugger as much as I do in C++ - and thus it doesn't matter as much - the debug symbols are necessary for line numbers in backtraces, which I *do* use/see quite often.

Do you have any numbers on how much turning this off improves the iteration time?

Pombal:
On my machine is saves 20s from a full build. But from your comment it seems like it's not a good idea so I'll close it.

PR #755, opened 2023-08-27
Disable lto in ci to improve compile times

Pombal:


Pombal:
@FredrikNoren Are user builds created from ci release builds or from somehwere else?

If so, should we create a new master configuration that enables fat lto?

ten3roberts:
Wouldn't setting rust flags in the CI override any rust flags we set in `.cargo/config.toml`?

That may be very unexpected and difficult to track down why a feature we add there does not appear in the build artifacts (but work locally).

Wouldn't perhaps a better approach be to create a cargo profile to avoid overriding all the rust flags already set?

Pombal:
What do you mean by a feature? This change should only affect performance.
I could create a master profile and disable lto in release but the performance hit is quite drastic in rust, unlike in C++ which is more moderate.

ten3roberts:
> What do you mean by a feature? This change should only affect performance. I could create a master profile and disable lto in release but the performance hit is quite drastic in rust, unlike in C++ which is more moderate.

Sorry, perhaps I wasn't as clear

*Disabling* lto is absolutely cgreat to reduce the compile time.

What I am concerned about is that we already set the `RUST_FLAGS` environment variable to things, and this (along with the debuginfo=0) would nullify all those existing flags.

What I mean by feature is a perhaps better worded as a "configuration"

As such, it is better to add there as a cargo *profile* rather than an env

https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

FredrikNoren:
@Pombal We're starting to set up nightly builds (or @chaosprint and @ten3roberts are), and those would be user builds. So for the CI I think it's fine if it doesn't have lto.

Pombal:
@ten3roberts It seems to me that a cargo profile is a more heavy handed way of doing it and also less explicit.

Not sure if it's the same in Rust but in games with C++ there's tipically a debug, release, profile and master configurations.
And it's important to minimize the number of them because the testing matrix, tooling changes, and congnitive overhead increases substantly when a new configuration is introduced.

I'm thinking that adding a new configuration to Rust will incur similar costs but I could be wrong.

Whereas changing lto directly in ci is very clear and explicit that this change only intends to trade performance for compile times in ci alone.
A new configuration could be used for other things and would have to be supported for other means. And so if we wanted to enable lto in ci again in the future it would be harder.

It's possible that these costs don't exist in rust and it's actually a good practice to have multiple configurations but as of now this rings alarm bells for me.

Pombal:
@FredrikNoren Thanks, I've enabled it again for nightly.

Pombal:
@ten3roberts I re-read what you said and I think I now understand your concern better.
The primary problem you mention is wiping out RUST_FLAGS. I agree this may lead to confusion.

So first off, this is already happening. Can we live with it until it becomes a problem? 
Second, is there another way to prevent this from happening without creating a new profile?

PR #760, opened 2023-08-28
Add public field to ambient.toml

FredrikNoren:


philpax:
Hm, maybe this should be an option when deploying a package for the first time, and then you can set it from the website? It seems like more of a deployment setting than a package setting, if that makes sense.

FredrikNoren:
@philpax Maybe we'll move it in the future but for now I'll just keep all the configuration in the ambient.toml, makes it much easier to reason about and work with. It's the same as other metadata (name, description, content etc.).

PR #772, opened 2023-08-31
Add ambient_version field to ambient.toml

FredrikNoren:


FredrikNoren:
@chaosprint I'm adding this so we can start preparing the website, but this field is what the ambient version manager should read from as well to figure out what version to download and run.

FredrikNoren:
@philpax Hm guess in the future we might want to support weaker version specifiers here too (like `^1.3`), to figure out what runtime version a package is compatible with

shared_crates/package/src/manifest.rs
```diff
@@ -73,6 +74,64 @@ fn return_true() -> bool {
...
+    Debug,
+    PartialEq,
+    parse_display::Display,
+    parse_display::FromStr,
+    serde_with::SerializeDisplay,
+    serde_with::DeserializeFromStr,
+)]
+pub enum AmbientRuntimeVersion {
+    #[display("{major}.{minor}.{patch}")]
+    Stable { major: u32, minor: u32, patch: u32 },
```
philpax, line 88: You might want to use `Version` here instead? You can then reuse its logic for semver etc

philpax:
lgtm but worth noting that the build metadata already includes the Ambient version it was built with. It might be useful to show both the supported and built versions on the website.

shared_crates/package/src/manifest.rs
FredrikNoren: @philpax Hm any reason we're not using https://crates.io/crates/semver instead of our own?

philpax: Oh, I didn't see this - the reason was to provide deterministic behaviour for comparisons, especially with our `-dev`/`-alpha`/etc editions, and it grew out of the existing data structure. It looks like `semver` should cover this, so it shouldn't be too problematic to replace it.

Issue #71, opened 2023-02-13
Add function to load a `World` with objects and remove `entity::spawn_template`


philpax:
```rust
let world = World::load_from_url("assets/Cube.glb/objects/main.json").await?;
// world.objects contains a dict of entityids to components
assert_eq!(world.objects.iter().next(), Some((EntityId(...), Components::new()...)));
/// can spawn individual objects, or spawn the entire world
let id = world.objects.get(&EntityId(...)).unwrap().spawn();
let id = world.spawn();
```

FredrikNoren:
`entity::spawn_template` has been removed

philpax:
Just realised this needs infrastructure on the guest side for the host to notify it of async events (unless we reuse `event::ENTITY_SPAWN`) for this. Maybe define another event like `event::WORLD_SPAWN`?

(What does `object_from_url` do for multiple entities in an object, anyway?)

FredrikNoren:
@philpax So there are two different things here; loading and spawning worlds has more to do with maps and other stuff. I think that probably has it's own flow, maybe something like `let map = World::load("url").await; map.spawn();` or something.

Then on the other hand we have `object_from_url`; this is for hot-reloading assets etc. In that case, the "base entity" is already created by the user, so to be notified when the rest of the entity is loaded we'd have to wait for some `object_loaded` component to be added to that entity (doesn't exist yet).

`object_from_url` doesn't handle multiple-entity objects right now, but it should handle it the same way as `model_from_url` does in the future; if it's one object it should attach the components to the entity; if it's multiple they should be created as children to it.

philpax:
`Prefab` is the term we use for a world being spawned in and updating/creating an entity or multiple entities. In line with that, we'd call this `Prefab::load_from_url("url")` and offer both the ability to spawn it immediately and the ability to poke around in its hierarchy before spawning it.

philpax:
A future extension of this could also be used to save entities for later loading. I'm leaving that as a future development as it'll have to tie in to our story around persistence and general I/O access.

PR #773, opened 2023-08-31
Guest world

### LINKED ISSUE #725: Add object-oriented `EntityId` methods
As game codebases scale, I begin to find this extremely noisy and hard to read:
```rs
entity::get_component(e, component());
entity::despawn_recursive(e);
entity::add_child(e, child);
```

I think that a much more readable (and idiomatic if you're looking at Ambient packages as objects) is this:
```rs
e.get(component());
e.despawn_recursive();
e.add_child(child);
```

philpax:
This is an attempt to switch the guest over to using a `World` struct instead of global functions to help resolve #725 and to enable #71 in the future.

However, there are a few blocking issues:
- The `AnimationNode` `Clone` and `Drop` implementations rely on modifying the ECS globally, which doesn't work when there's a `World` floating around. I've disabled this for now just to make it compile, but that's not a workable solution.
- Anything involving callbacks requires passing the `World` through, including event subscriptions.
- Every bit of Ambient code will need to be updated.

I'm leaving this as a draft PR to pick up later, but it seems like something we should do at some point.


FredrikNoren:
@philpax Hm yeah.. I'm not entirely happy with AnimationNode, and it feels like this might actually expose that the design isn't that good. I'll take a look if I can figure out how to do the animation stuff differently

philpax:
With the recent animation API changes, this is now viable again. I'll revisit once I'm done with other work.

PR #774, opened 2023-08-31
Expose inactivity shutdown

pierd:
I've noticed that we already have one. Fixing it and exposing as an argument so that it can be easily used on the backend.

### DIFF
--- a/app/src/cli/mod.rs
+++ b/app/src/cli/mod.rs
@@ -234,6 +234,10 @@ pub struct HostCli {
     /// Private key for the certificate
     #[arg(long)]
     pub key: Option<PathBuf>,
+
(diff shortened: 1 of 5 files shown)


philpax (commented):
What's the reason for exposing the join handle from the server?

pierd:
> What's the reason for exposing the join handle from the server?

We spawn the server in a separate task so we have to wait on something. For `run` it blocks on client, for `serve` we just waited for ctrl+c no matter what is happening to the server. For the process to actually end when the server ends I have to wait on its task directly.

pierd:
@philpax what do you think about using an explicit struct instead of a tuple? For example `ServerHandle`?

philpax:
Sounds good to me!

PR #780, opened 2023-09-01
chore: update msrv to 1.70.0

philpax:


ten3roberts:
How come this wasn't caught in the CI?

Doesn't `rust-version` force a build for that rust version, or is that only for non-workspaces?

philpax:
We misunderstood it - it doesn't set the current version, it's just a minimum version constraint. The CI builds with whatever the latest version is, but checks if it's greater than the set rust-version.

We can manually check for the MSRV in the CI if we want.

PR #781, opened 2023-09-01
Add unity asset loading example

FredrikNoren:


guest/rust/examples/assets/unity/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "unity"
+
+edition = "2021"
+publish = false
+rust-version = { workspace = true }
+version = { workspace = true }
+
+[dependencies]
+ambient_api = { path = "../../../api" }
```
philpax, line 10: workspace?

guest/rust/examples/assets/unity/download.sh
```diff
@@ -0,0 +1,37 @@
+export BASE="${EXAMPLES_ASSETS_HOST}/unity/Dynamic%20Nature%20-%20Mountain%20Tree%20Pack"
```
philpax, line 1: Is EXAMPLES_ASSETS_HOST defined somewhere?

FredrikNoren: Yeah it's a github secret, because it's not public content

FredrikNoren: (In the future we should switch to public content, but to get started I wanted to use something I know at least works somewhat)

PR #782, opened 2023-09-03
Add support for Sentry native crashes and allow users to disable it or use their own DSN with client settings.

Pombal:
Upload debug symbols for nightly builds.
Add documentation for Ambient settings.

### DIFF
--- a/crates/settings/src/general.rs
+++ b/crates/settings/src/general.rs
@@ -1,0 +1,30 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Default, Debug)]
+pub struct GeneralSettings {
+    pub api_token: Option<String>,
+    pub sentry: Sentry,
+}
+
+impl GeneralSettings {
+    pub fn is_sentry_enabled(&self) -> bool {
+        self.sentry.0 .0
+    }
+
+    pub fn sentry_dsn(&self) -> String {
+        self.sentry.0 .1.clone()
+    }
+}
+
+// enabled, dsn
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Sentry(pub (bool, String));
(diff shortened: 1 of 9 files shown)


FredrikNoren (approved):
Nice!
For the settings doc to appear you need to update the SUMMARY.md file as well. 
Also there's a .github/release workflow which will also need the debug symbols (which we use to release point versions of ambient)

Pombal:
@FredrikNoren Good catch!
I've updated the docs and moved the symbol upload to .github/release. I thought we were going to do releases from nightly builds and missed release.yml.

crates/settings/src/general.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    pub fn sentry_dsn(&self) -> String {
+        self.sentry.0 .1.clone()
+    }
+}
+
+// enabled, dsn
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Sentry(pub (bool, String));
```
philpax, line 21: I'd use explicit fields for this to make it obvious what they represent, especially for people editing the config

docs/src/user/settings.md
```diff
@@ -0,0 +1,14 @@
+# Settings
+
+Ambient supports a number of settings that can be configured using the `settings.toml` file. This file is located under the platform's config directory, e.g. `C:\Users\*USER*\AppData\Roaming\Ambient\Ambient\config\settings.toml` on Windows.
+
+## Settings
+
+[general]\
```
philpax, line 7: You can use three-backticks followed by toml here, like here https://github.com/AmbientRun/Ambient/blob/main/docs/src/reference/models.md?plain=1#L8-L12

philpax (commented):
LGTM, but a few minor tweaks before merge

PR #783, opened 2023-09-04
Fix campfire deploy-all call in deploy-packages action

pierd:


philpax:
Ah, you've beaten me to it! Note that the changes need to be committed, because `deploy-all` modifies the packages with their deployment IDs - are you planning on tackling that as well?

pierd:
> Ah, you've beaten me to it! Note that the changes need to be committed, because `deploy-all` modifies the packages with their deployment IDs - are you planning on tackling that as well?

Yeah, I'll do it

philpax (approved):

PR #798, opened 2023-09-05
feat(physics): add optional range to raycasts

philpax:
This is a fairly uncontroversial change, except for the part where it doesn't actually work - the `physics` example doesn't produce any raycast results with a set distance, even when the cube is obviously within 5 -Z units of the ground.

I tried chucking in a `dbg!` to see what the range being passed into the raycast was:
image
but that results in the runtime crashing on startup:
```
[2023-09-05T16:55:26Z INFO  ambient_audio::stream] Audio stream config: StreamConfig { channels: 2, sample_rate: SampleRate(48000), buffer_size: Default }
[2023-09-05T16:55:26Z INFO  minidumper::ipc::server] captured minidump
```

@FredrikNoren did you use `f32::MAX` for a specific reason? Any FFI nasal demons I should be aware of?


FredrikNoren:
@philpax Hm strange. Don't think there's any FFI thing, think the c++ physx examples just use MAX too

FredrikNoren:
@philpax Wait I think I get the minidump thing on all startups, even without this change

PR #800, opened 2023-09-06
Fix disconnecting before entering the name in AFPS

pierd:


pierd:
FYI @chaosprint

philpax (approved):
I was confused at first, but then I remembered `wait_for_component` will *also* return if the entity was deleted. Maybe we should make that more obvious somehow.

pierd:
> I was confused at first, but then I remembered `wait_for_component` will _also_ return if the entity was deleted. Maybe we should make that more obvious somehow.

A bit heavy handed but we could add `#[must_use]` to `wait_for_component`.

philpax:
> > I was confused at first, but then I remembered `wait_for_component` will _also_ return if the entity was deleted. Maybe we should make that more obvious somehow.
> 
> A bit heavy handed but we could add `#[must_use]` to `wait_for_component`.

You know that's not actually a bad idea, I'll do that

PR #819, opened 2023-09-08
Make all hooks free functions

FredrikNoren:
Standardize so everything is `use_` for hooks.


philpax:
Might also be a good idea to rename to `cx: Context`?

FredrikNoren:
@philpax Yup

PR #824, opened 2023-09-08
fix: deploy web

ten3roberts:


philpax (approved):
there has got to be a way to reuse these steps

ten3roberts:
> there has got to be a way to reuse these steps

Agreed, I tried to mitigate it by using campfire, so there is that.

Though I believe this conversation came up yesterday as well in another PR, don't quite recall the conclusion of it

Issue #766, opened 2023-08-29
Write a style guide for error/user-facing messages


philpax:
As we have more and more user-facing text (especially in the more "compiler-y" aspects of the runtime and API), I've realised that the text is not very consistent in style, which can be quite jarring to read when you have multiple messages in close proximity.

I'm thinking about writing a style guide and then updating some of the more prominent strings to match that guide.

Here's a first pass at some of the guidelines:
- Use American English. (Sadly, the Americans won the race to standardise technological terms.)
- Messages should start with some consistent casing. Not sure if lower-case or upper-case.
- Objects (packages, components, etc) should be referred to with surrounding backticks. That is, ``the package `bobcat` does not have a component `is_bobcat` ``.
- Paths should be referred to with quotation marks surrounding them. (The Rust debug implementation for paths does this automatically.)
- The additional context (as in the anyhow kind) for an error should be a message in itself, not just a fragment. That is, `Error while processing single pipeline in "lol.toml"`, instead of `In pipeline "lol.toml"`.
- In general, try to optimize for easy copy-ability / clicking. If it's a link, you should be able to click it in your terminal without having to manually select it. 
- Use logging instead of `println` within the application (`app/` and elsewhere). This is the only one I'm not sure about, but it's what we're currently doing, and it gives us timestamps.

ten3roberts:
> Use logging instead of println within the application (app/ and elsewhere). This is the only one I'm not sure about, but it's what we're currently doing, and it gives us timestamps.

Yes, not entirely sure about that one either. On one hand logging provides a nice UI for what is going on, on another hand logging is intended for process introspection with fields, location etc which is not relevant for user facing status, but rather a background thing you can enable in tangent to the user UI

pierd:
IMHO logging > println

remembering that everything INFO and above is user facing (DEBUG/TRACE less so)

pierd:
> * Use American English. (Sadly, the Americans won the race to standardise technological terms.)

Agreed 😢 also "standardize"

![bitmoji](https://sdk.bitmoji.com/render/panel/d6dac8b7-7746-40bb-89d1-1c9f85dd32a0-77c97d43-89a9-4409-8cdf-477faad3f1a4-v1.png?transparent=1&palette=1&width=246)

philpax:
[...]

PR #826, opened 2023-09-08
Moved code for fog example into standalone modules

droqen:
Experimentally reorganized the fog example's code into four modules
The purpose of this is to make it easier to comprehend and edit the code at a high level at a glance, see how the first few lines of code the user sees will help them understand what it is doing immediately:
```
[..]
#[main]
fn main() {
    default_camera::init();
    ground_and_cubes::init();
    let sun = fog_lighting::spawn_sun();
    ui_sliders_for_sun::init(sun);
}
```
It is also to make the code more re-usable despite the requirement of complex `use` statements; I might want to use the 'fog' from the fog example in my own package, and the encapsulation of all fog-related components in one place, exclusive of all other components (for e.g. the many camera components), means I can copy the `fog_lighting` module into my own code with no changes and begin using it immediately:
```
mod fog_lighting {
    use ambient_api::{
        core::{
            app::components::main_scene,
            rendering::components::{
                fog_color, fog_density, fog_height_falloff, light_diffuse, sky, sun,
            },
            transform::{components::rotation, concepts::make_transformable},
        },
        prelude::*,
    };
    pub fn spawn_sun() -> EntityId {
        let sun = Entity::new()
            .with_merge(make_transformable())
            .with(sun(), 0.0)
            .with(rotation(), Quat::from_rotation_y(-1.))
            .with(main_scene(), ())
            .with(light_diffuse(), Vec3::ONE)
            .with(fog_color(), vec3(1., 1., 1.))
            .with(fog_density(), 0.1)
            .with(fog_height_falloff(), 0.01)
            .spawn();

        Entity::new()
            .with_merge(make_transformable())
            .with(sky(), ())
            .spawn();

        sun
    }
}
```

### DIFF
--- a/guest/rust/examples/rendering/fog/src/client.rs
+++ b/guest/rust/examples/rendering/fog/src/client.rs
@@ -1,83 +1,127 @@
-use ambient_api::{
-    core::{
-        app::components::main_scene,
-        camera::{
-            components::{aspect_ratio_from_window, fog},
-            concepts::make_perspective_infinite_reverse_camera,
-        },
-        primitives::components::{cube, quad},
-        rendering::components::{
-            cast_shadows, color, fog_color, fog_density, fog_height_falloff, light_diffuse, sky,
-            sun,
-        },
-        transform::{
-            components::{lookat_target, rotation, scale, translation},
-            concepts::make_transformable,
-        },
-    },
-    prelude::*,
-};
+use ambient_api::prelude::*;
 
 #[main]
 fn main() {
-    Entity::new()
-        .with_merge(make_perspective_infinite_reverse_camera())
-        .with(aspect_ratio_from_window(), EntityId::resources())
-        .with(main_scene(), ())
-        .with(fog(), ())
-        .with(translation(), vec3(0., -5., 3.))
-        .with(lookat_target(), vec3(0., 0., 2.))
-        .spawn();
-
-    let sun = Entity::new()
-        .with_merge(make_transformable())
-        .with(sun(), 0.0)
-        .with(rotation(), Quat::from_rotation_y(-1.))
-        .with(main_scene(), ())
-        .with(light_diffuse(), Vec3::ONE)
-        .with(fog_color(), vec3(1., 1., 1.))
-        .with(fog_density(), 0.1)
-        .with(fog_height_falloff(), 0.01)
-        .spawn();
-
-    Entity::new()
-        .with_merge(make_transformable())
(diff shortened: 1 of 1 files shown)


droqen:
My first Ambient PR so I have requested a couple reviews! 👼
A very simple one, but it may have some bigger implications style-wise. (Not directly related but #766-adjacent?)

philpax:
Good call on splitting them up, but I think the modules might be a bit much (especially because you have to re-import everything)

How does it look if you split them up into functions in the same module? I think you could get the same benefit of modularity, but with less duplication.

It might also be worth noting that you can do imports within a function:

```rust
fn lol() {
  use ambient_api::core::apps::components::main_scene;
  // do something with `main_scene`
}
```

FredrikNoren:
I agree with @philpax, the module in module approach seems much more verbose than necessary. Also not sure if you're using this but with VSCode you can easily manage imports by pressing command-dot on anything that is missing an import to include it (not sure what it is in Windows, maybe ctrl-dot), and equivalently it can remove unused imports for you if you click an unused import and press cmd-dot.

droqen:
@philpax What do you mean by "especially because you have to re-import everything"?

Also I did _not_ know you can import within a function :o

FredrikNoren:
@droqen Hm I think imports in functions also has the problem that the code just becomes unnecessarily long. You'll have a lot of duplicate imports in the same file since they're imported for each function.

> Cleaning up unused imports all at once --> Maybe there is a command or package for doing this

Ctrl+. when it's yellow should show a menu for this. It'll only do it per line which is a bit unfortunate (for typescript there's an option to do it for all imports).

> Ctrl+. does not work for my packages' components --> As I have come to understand it, this may actually a problem with my config?

That should work. Can you try just creating a new "clean" ambient project with `ambient new` and see if it works there?

> Ctrl+. does not work for remote packages' components --> I don't think this is possible yet, but maybe I underestimate the magic

Think that should also work, right @philpax?

philpax:
Yeah, should just work - it's all part of the `packages` module, which your code has access to.

droqen:
image

image

droqen:
Autocompleting 'packages' does not work:

image

Some things work by VS Code 'guessing' at me possibly wanting to type a word that I've seen before, but it has no context so it autocompletes things in wrong places:

image

philpax:
Yeah, that should definitely work. We should figure out what's going on with your setup at some point...

Issue #843, opened 2023-09-12
A general solution to client / server ECS divergence


philpax:
While debugging an issue this morning, @FredrikNoren and I realised there's a much bigger issue - the client and the server can disagree on the state of a component/entity, and there is no general way to unify them.

The issue we saw was:
- a package attaches `model_from_url` to a player on the server
- `model_from_url` runs on the client, attaching some children to the player entity
- another package runs on the server, adding a `head` entity as a child to the *server* children of the player, which synchronises to the client, overriding the children set by `model_from_url`

There are workarounds for this particular issue, and we'll likely document for them for this immediate release. For this example, we'll move the model to a separate entity so that its children are not interfered with by the server. For Tangent, we'll wait for the model load on the client before attaching the HUD.

However, the general problem extends beyond this. This could happen to any component where the client updates the ECS, and then the server replicates a newer version, replacing the client's version.

There are a couple of solutions to this, each with their own trade-offs:

1) have three versions of the same component, and then have a system to unify them. For children, this would be `server_children`, `client_children_delta`, and `authoritative_children`, where SC and AC are the same on the server, and AC is SC + CCD on the client.
2) have an unnetworked canonical component on both sides, and then an invisible networked component whose job it is to provide the "base state". On the server, there's one `children` component that is canonical and unnetworked, then there's a `server_children` component that is networked and not visible to the guest that gets synced from `children`. On the client, `children` is automatically updated by a system to be a composition of `server_children` and `client_children_delta`. Updating `children` on the client will update CCD for you.
3) An OverlayFS-like solution where you always have access to the base server world on the client, but the client has an "overlay" on top of the server world that it uses for all rendering/etc. When you update the client world, the change you've made is compared to the server state, and the delta is stored in the overlay, so that it can be reapplied should the server state change. The behaviour of the delta is user-specified and per-component.
  a) A variation of this where it's more explicit, and you can create your own overlays and choose what the runtime renders with/operates over.

All of these complicate the data model in different ways, but this is not necessarily avoidable complexity. The problem of the client wanting to make some changes to the state that the server will always exist, whether it's for prediction, animation, client-only entities, or more.

Open questions:
- What does it mean to remove something that the server spawned? Both entities and attached components.
- Does one delta behaviour for each component cover all use cases? For translation, you might want the client to override the position at all times, but you may also want it to be applied as a delta to the server position.
- How will this work when there are multiple packages interacting with the same component?

philpax:
[...]

Issue #230, opened 2023-03-15
ECS API function changes


philpax:
As part of the UI port, @FredrikNoren noticed that the ECS API on the guest isn't aligned with the host's. This is mostly fine (especially with the guest bridge), but there are a couple of things that we should address:

- All ECS operations are currently infallible and will panic if they fail. They should instead return a result, so that guest code can decide what to do if a component doesn't exist.
- Consider supporting `entity::get_component_ref`/`Entity::get_ref` that allows retrieval of `&T` where it's safe to convert between the `wit-bindgen` inner value and `T` (all primitives, all Rust-defined types)
- Consider laying the groundwork for other `World`s at some point? Not sure what this would look like as WASM modules currently belong to worlds in themselves

PR #845, opened 2023-09-12
Derived `children` from `parent`

FredrikNoren:
This fixes client-server synchronization of hierarchies; instead of synchronizing the `children` we now only synchronize `parent` and instead construct `children` based on parents. (Re. #843 )


docs/src/reference/hierarchies.md
```diff
@@ -88,3 +84,9 @@ mesh_to_world = local_to_world * mesh_to_local
 This also means that you can attach a mesh in the middle of a hierarchy, with an offset. For instance, if you have
 a bone hierarchy on a character, you can attach an mesh to the upper arm bone, but without `mesh_to_local/world` it
 would be rendered at the center of the arm (inside the arm), so by using `mesh_to_local/world` you can offset it.
+
+## Opting out of automatically derived children
+
+If you whish to manage the `children` component yourself, you can attach an `unmanaged_children` component to your
```
philpax, line 90: wish

guest/rust/api_core/src/entity.rs
```diff
@@ -224,20 +225,24 @@ pub fn mutate_component_with_default<T: SupportedValue + Clone + PartialEq>(
...
 }
 
 /// Removes `child` as a child to `entity`.
 pub fn remove_child(entity: EntityId, child: EntityId) {
-    if has_component(entity, children()) {
-        mutate_component(entity, children(), |children| {
-            children.retain(|x| *x != child)
-        });
+    if has_component(entity, unmanaged_children()) {
+        if has_component(entity, children()) {
```
philpax, line 241: this `if` not necessary, `mutate_component` does nothing if it doesn't have that component

philpax (approved):

guest/rust/api_core/src/entity.rs
FredrikNoren: @philpax Hm yeah, it feels like the mutate_component etc. should return Results, so it's up to the user to handle situations like that? Then it could just be `.ok()` on it to ignore the result.

philpax: Yeah, that's the plan with #230 / #773, but `mutate_component` already returns an `Option<T>` with the resulting value; you can check for `None` if you want

Issue #835, opened 2023-09-11
Ambient recompiles from `ambient_native_std` upwards every single time


philpax:
In the last few days, `cargo run` appears to need to recompile Ambient from `ambient_native_std` upwards, resulting in a miserable dev experience with a 30s wait each time at minimum.

I believe this may have been introduced with the merge of the web client, but we need to check. I have no idea what would cause this - is something being marked as dirty somewhere?

FredrikNoren:
Hmm could it be because of the git_revision thing in there? git_revision_full

FredrikNoren:
@pierd ^

pierd:
I'll have a look if it's that or if I can spot anything else that's sus

pierd:
@FredrikNoren Yip. `git_version!` macro is the culprit. I'm looking for a better way to solve this.

PR #851, opened 2023-09-13
Fix excessive rebuilds

pierd:
https://github.com/AmbientRun/Ambient/issues/835

This works almost like the git-version macro, the main difference being: it keeps only the `.git/logs/HEAD` as a dependency for rebuilds while git-version also tracks `.git/index`.


philpax (approved):
LGTM, does it resolve the issue?

pierd:
> LGTM, does it resolve the issue?

It does. I'll do some more testing on Windows.

PR #852, opened 2023-09-13
Disable watch by default and add `--watch` to `campfire web serve`

ten3roberts:


ten3roberts:
Watching is good to have you lazily build something, but due to the constant rebuilding and limited compute power it may be good to not build the web client on save if you are writing things deep in the code depedency tree

campfire/src/web/serve.rs
```diff
@@ -31,13 +34,19 @@ impl Serve {
                 .context("Failed to run npm install")?;
         }
 
+        self.build
```
ten3roberts, line 37: Ensures the package is always built before running the web server, to avoid situations where you are serving a `pkg` that has been hanging around and wondering why your changes don't show.

That has happened a few of us now 😝

philpax (approved):

PR #858, opened 2023-09-13
feat: initialize wasm in parallel

ten3roberts:
This improves the launch speed of games using multiple packages


philpax:
Just to confirm how this works - each module is instantiated in a Tokio task, which is run on a worker thread somewhere, and then `block_in_place` is used to dedicate that worker thread to the task of instantiation so that other threads can process the other tasks. Is that right?

That seems reasonable to me; is there any risk of all of the worker threads being used on instantiation and blocking all async work in the meantime? My understanding was that you should generally use threads for compute-heavy work to avoid stalling the executor, which is why I explicitly used threads before.

philpax (approved):

Issue #651, opened 2023-08-08
Rendering on Web on Windows is broken


FredrikNoren:
Something seems to be wrong on chrome on Windows with the renderer on the web

ten3roberts:
Yes. @Pombal discovered this as well, and it has been added to my task list.

Thansk

ten3roberts:
Though what exactly does *instance* rendering mean in this context?

Is it referring to the indirect primitives, or the `instancing` example, or something else. Afaik we don't really make use of instancing in the common sense

FredrikNoren:
@ten3roberts I just meant the web version of ambient

Pombal:
I narrowed it down to an issue with the Fxc shader compiler. I have side by side captures of a good and a broken build but debugging the shader assembly gets tricky because we're not comparing apples to apples, it's DXBC vs DXIL.

To repro natively change the backend to wgpu::Backends::DX12 with wgpu::Dx12Compiler::Fxc in gpu.rs.

philpax:
For some more data, Tangent on the web on Windows works, but `a_foggy_scene` does not: it looks like this 
![image](https://github.com/AmbientRun/Ambient/assets/707827/6bb4e6a3-9d97-488e-ad37-272b38852655)

PR #868, opened 2023-09-14
Fix windows rendering

ten3roberts:
Fixes: #651


ten3roberts:
Fixes: #651

.vscode/settings.json
```diff
@@ -1,8 +1,9 @@
 {
-  "editor.detectIndentation": false,
-  "editor.tabSize": 4,
```
ten3roberts, line 3: Ironically, this was indented with only 2 spaces

crates/renderer/src/tree_renderer.rs
```diff
@@ -558,63 +556,66 @@ impl TreeRenderer {
...
+                } else if self.config.render_mode == RenderMode::Direct {
+                    for (i, &(id, primitive_idx)) in mat.primitives.iter().enumerate() {
+                        let primitive = &world.get_ref(id, primitives()).unwrap()[primitive_idx];
+
+                        let mesh = mesh_buffer.get_mesh_metadata(&primitive.mesh);
+                        let index = offset + i as u64;
+
+                        render_pass.draw_indexed(
+                            mesh.index_offset..(mesh.index_offset + mesh.index_count),
+                            // mesh.base_offset as i32,
```
ten3roberts, line 601: Strange.

Adding `mesh.base_offset` makes the direct render mode work on the web, but stops it working on native (gray screen).

Doing 0 yields the other way around

🤔

ten3roberts: Update:

This line fixed direct rendering on windows web, but breaks direct rendering on mac web, as well as native

ten3roberts:
Something odd is going on with DirectX12.

Found a temporary work around so will use that for the time being

PR #874, opened 2023-09-15
Fix new package api version. Fixes #873

### LINKED ISSUE #873: `ambient new` template doesn't work with nightly
Repro:

1. `ambient new hello`
2. `cd hello`
3. `ambient run`

Gives the following error:

```sh
Error: Failed to build Rust in "/Users/fredrik/github/testy23/build/testy_23"

Caused by:
    Failed to compile
    stderr:
        Updating crates.io index
    error: failed to select a version for the requirement `ambient_api = "^0.3.0-nightly-2023-09-15"`
    candidate versions found which didn't match: 0.2.1, 0.2.0, 0.1.1, ...
    location searched: crates.io index
    required by package `testy_23 v0.0.1 (/Users/fredrik/github/testy23)`
    perhaps a crate was updated and forgotten to be re-vendored?
```

FredrikNoren:
@pierd @philpax Can you just verify this logic makes sense? (I just removed the production config and made it always use the not-production one)

Resolves #873

### DIFF
--- a/app/src/cli/new_package.rs
+++ b/app/src/cli/new_package.rs
@@ -48,7 +48,11 @@ pub(crate) fn handle(
     // Write the files to disk.
     let ambient_toml = include_str!("new_package_template/ambient.toml")
         .replace("{{id}}", id.as_str())
-        .replace("{{name}}", name);
+        .replace("{{name}}", name)
+        .replace(
+            "{{ambient_version}}",
+            &format!("{}", ambient_native_std::ambient_version().version),
+        );
 
     let cargo_toml = {
         // Special-case creating an example in guest/rust/examples so that it "Just Works".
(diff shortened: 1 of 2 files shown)


app/src/cli/new_package.rs
```diff
@@ -48,7 +48,11 @@ pub(crate) fn handle(
     // Write the files to disk.
     let ambient_toml = include_str!("new_package_template/ambient.toml")
         .replace("{{id}}", id.as_str())
-        .replace("{{name}}", name);
+        .replace("{{name}}", name)
+        .replace(
+            "{{ambient_version}}",
+            &format!("{}", ambient_native_std::ambient_version().version),
```
philpax, line 54: Maybe move out this `ambient_version` call upwards and use it for both this and `cargo_toml`?

philpax (approved):
Seems reasonable to me.

Issue #534, opened 2023-06-27
Collider doesn't inherit transform from parent


droqen:
![image](https://github.com/AmbientRun/Ambient/assets/5104632/bd211b5a-8976-4e99-8f8c-050bcd9e0d2f)

You can use this minimal repro code (server.rs):

```
use ambient_api::{
    components::core::{
        app::main_scene,
        camera::aspect_ratio_from_window,
        ecs::{children,parent,},
        physics::{cube_collider,visualize_collider,},
        primitives::{cube,quad,},
        transform::{lookat_target, translation, local_to_parent, local_to_world, },
    },
    concepts::{make_perspective_infinite_reverse_camera, make_transformable},
    prelude::*,
};

#[main]
pub fn main() {

    Entity::new()
        .with_merge(make_perspective_infinite_reverse_camera())
        .with(aspect_ratio_from_window(), EntityId::resources())
        .with_default(main_scene())
        .with(translation(), Vec3::ONE * 5.)
        .with(lookat_target(), vec3(0., 0., 0.))
        .spawn();

    let parent_position_only = Entity::new()
        .with(name(), "position only (parent)".to_string())
        .with_merge(make_transformable())
        .with(translation(), vec3(-2., 2., 0.))
        .with_default(local_to_world())
        .spawn();

    let child_mesh_and_collider = Entity::new()
        .with(name(), "mesh and collider (child)".to_string())
        .with_default(cube())
        .with(cube_collider(), Vec3::ONE)
        .with_default(visualize_collider())
        .with_default(local_to_parent())
        .spawn();

    entity::add_component(parent_position_only, children(), vec![child_mesh_and_collider]);
    entity::add_component(child_mesh_and_collider, parent(), parent_position_only);

}
```

A slightly more detailed implementation in latest push to my working repo here:

https://github.com/droqen/23-07Buck/blob/master/test02-parent-collider/src/server.rs

philpax:
Hm, I think this might be one for @FredrikNoren. The issue is that the ECS <-> physics world sync systems work solely on changes to translation/rotation/scale, and don't account for the hierarchy. I'm not sure what the correct fix would be here, especially because there are nuances around *how* the hierarchy should be represented in the physics world.

I can think of five cases:
1. non-physics object is moved around, and a physics object is a child of it
    a. Does the physics child get teleported, or is a force applied to move it to its new absolute position?
2. non-physics object stays still, but a physics object that is the child of it moves around
[...]

PR #885, opened 2023-09-17
Fix #803 child collision volumes

### LINKED ISSUE #803: Child entity collision volumes ?
I'm trying to add collision detection to the following scene.

I have the ground entity:
``` rust
    Entity::new()
        .with(name(), "ground".to_string())
        .with_merge(make_transformable())
        .with(plane_collider(), ())
        .with(visualize_collider(), ())
        .with(scale(), Vec3::ONE * 200.)
        .with(quad(), Default::default())
        .with(color(), vec4(0.2, 1., 0., 1.))
        .with(
            pbr_material_from_url(),
            assets::url("pipeline.toml/1/mat.json"),
        )
        .spawn();
```

I then have the ship entity as a child of a hierchy:

``` rust
            let player_mesh_id = Entity::new()
                .with(name(), "PlayerMesh".to_string())
                .with(parent(), player_controller_id)
                .with(local_to_parent(), Default::default())
                .with(local_to_world(), Default::default())
                .with_merge(make_transformable())
                .with(prefab_from_url(), assets::url("fighter01.glb"))
                .with(dynamic(), true)
                .with(sphere_collider(), 1.0)
                .with(visualize_collider(), ())
                .with(scale(), Vec3::ONE * 0.3)
                // rotate blender mesh to fit world coordinates
                .with(
                    rotation(),
                    Quat::from_rotation_z(-std::f32::consts::FRAC_PI_2)
                        * Quat::from_rotation_x(-std::f32::consts::FRAC_PI_2),
                )
                .spawn();
```

If we render the scene it looks like this:
image        

The `visualize_collider` rendering is strange. The red lines on the camera perspective are from the `sphere_collider` on the `PlayerMesh` entity. 

You can see a small helper widget rendering for the ground plane but there's no further outline of the plane.

I have the collision message handler setup like
``` rust
    Collision::subscribe(move |msg| {
        println!("Bonk! {:?} collided", msg);
    });
```

But nothing is being triggered. I would expect when the ship mesh collides with the ground plane it would trigger....

for further context here is the whole hierarchy of the player.

``` rust
   spawn_query(is_player()).bind(move |players| {
        for (player_id, _) in players {
            let player_controller_id = Entity::new()
                .with(name(), "PlayerController".to_string())
                .with(parent(), player_id)
                .with(local_to_parent(), Default::default())
                .with(local_to_world(), Default::default())
                .with(inv_local_to_world(), Default::default())
                .with_merge(make_transformable())
                .with(translation(), Default::default())
                .with(rotation(), Default::default())
                .with(player_controller_offset_lerp_end(), Vec3::ZERO)
                .with(player_controller_offset_lerp_t(), 0.0)
                .with(player_controller_strafe_direction(), 0.0)
                .with(player_controller_offset_rot_end(), Quat::IDENTITY)
                .spawn();

            let player_mesh_id = Entity::new()
                .with(name(), "PlayerMesh".to_string())
                .with(parent(), player_controller_id)
                .with(local_to_parent(), Default::default())
                .with(local_to_world(), Default::default())
                .with_merge(make_transformable())
                .with(prefab_from_url(), assets::url("fighter01.glb"))
                .with(dynamic(), true)
                .with(sphere_collider(), 1.0)
                .with(visualize_collider(), ())
                .with(scale(), Vec3::ONE * 0.3)
                // rotate blender mesh to fit world coordinates
                .with(
                    rotation(),
                    Quat::from_rotation_z(-std::f32::consts::FRAC_PI_2)
                        * Quat::from_rotation_x(-std::f32::consts::FRAC_PI_2),
                )
                .spawn();

            entity::add_child(player_controller_id, player_mesh_id);
            entity::add_components(
                player_id,
                Entity::new()
                    .with(name(), format!("player-root-{}", player_id).to_string())
                    .with(children(), vec![player_controller_id])
                    .with(player_controller_ref(), player_controller_id)
                    .with(player_heading_vec(), Vec3::X)
                    .with(player_heading_vec_lerp_end(), Vec3::ZERO)
                    .with(current_pitch_angle(), -20.0)
                    .with(current_yaw_angle(), 0.0)
                    .with(player_speed(), default_speed)
                    .with(player_mouse_bounds(), Vec4::ONE)
                    .with(player_hud_trackable(), Default::default())
                    .with(local_to_world(), Default::default())
                    .with(is_parked(), false)
                    .with_merge(make_transformable())
                    .with(translation(), Vec3::Z * 5.0), // offset above the ground
            );
        }
    });
```

kevzettler:
This PR addresses #803

This PR adds a new query to the physx system to update entities that have `physics_shape` components and their `local_to_world` changes. This covers the case where an entity is the child of a hierarchy and has a physics collider attached. The child entities may not change its `translation` or `rotation` components but its parents will and the child needs to update its collider state accordingly.

Heres an example gif with a `visualize_collider` component. The collider is generate from the asset pipeline and the ship model is a child of a parent entity that handles all the movement and rotation.
collider

I need feedback and guidance on this. I basically copied this from the `query((translation().changed(), rotation().changed())).incl(physics_shape());` query implementation. Let me know any thoughts on how to improve.

### DIFF
--- a/crates/physics/src/physx.rs
+++ b/crates/physics/src/physx.rs
@@ -195,6 +195,7 @@ pub fn sync_ecs_physics() -> SystemGroup {
                         let is_kinematic = world.has_component(id, kinematic());
                         let (_scale, rot, pos) = localworld.to_scale_rotation_translation();
                         if let Ok(body) = world.get(id, rigid_dynamic()) {
+                            //                            println!("child physx is dynamic {:?} ", id);
                             let pose = PxTransform::new(pos, rot);
                             if is_kinematic {
                                 body.set_kinematic_target(&pose);
@@ -204,8 +205,10 @@ pub fn sync_ecs_physics() -> SystemGroup {
                                 body.set_angular_velocity(Vec3::ZERO, true);
                             }
                         } else if let Ok(body) = world.get(id, rigid_static()) {
+                            //                            println!("child physx is static {:?} ", id);
                             body.set_global_pose(&PxTransform::new(pos, rot), true);
                         } else if let Ok(shape) = world.get_ref(id, physics_shape()) {
+                            //                            println!("child physx has a shape? {:?} ", id);
                             let actor = shape.get_actor().unwrap();
                             let pose = PxTransform::new(pos, rot);
                             if !is_kinematic {
--- a/crates/physics/src/physx.rs
+++ b/crates/physics/src/physx.rs
@@ -5,7 +5,7 @@ use ambient_core::{
     transform::{local_to_world, rotation, scale, translation},
 };
 use ambient_ecs::{
-    components, ensure_has_component, parent, query, FnSystem, QueryState, Resource, SystemGroup,
+    components, ensure_has_component, query, FnSystem, QueryState, Resource, SystemGroup,
 };
 use ambient_native_std::asset_cache::SyncAssetKey;
 use glam::{EulerRot, Quat, Vec3};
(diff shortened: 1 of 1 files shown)


crates/physics/src/physx.rs
```diff
@@ -99,6 +99,13 @@ pub fn sync_ecs_physics() -> SystemGroup {
         query((translation().changed(), rotation().changed())).incl(physics_shape());
     let translation_rotation_q2 = translation_rotation_q.query.clone();
 
+    // we need a seperate query here for child entities
+    // Though they have rotation and translation those values may not update
+    // the parents will and local_to_world.changed() will need to be updated
```
FredrikNoren, line 104: Hm... it might make more sense to just replace the translation_rotation one, and then maybe have something that reads back changes to ltw to translation rotation.

In general we need to look over the physics system, it was built a long time ago for slightly different requirements that what's needed now

philpax: How do you feel about merging this as-is and then fixing it up later? The child colliders have been an issue for some time (https://github.com/AmbientRun/Ambient/issues/534), so I think unblocking this and revisiting it later is a decent trade.

FredrikNoren: I think as is it will conflict with the translation_rotation one. @kevzettler if you add `.incl(local_to_parent())` on the query (line 106) it should only affect entities that are children.

Also line 107 it's cloning the wrong query.

But in general; Let's revisit this once the launch is out; I'm not sure how else this will affect things so not sure if anything will break by doing this. We need to set up more test cases for physics also.

kevzettler: @FredrikNoren thanks for the code review I have made the fixes in https://github.com/AmbientRun/Ambient/pull/885/commits/9af616c7468bccf29590bbc46f0947ece02d3939

I will look into adding test cases but I do not want to make a large investment if this physics implementation gets rewritten. I will work to keep this branch up todate with main because I will need this child collision behavior for my project.

FredrikNoren: @kevzettler Great! With those fixes I think it actually can't affect anything else and it looks good to me so I'll go ahead and merge it

PR #901, opened 2023-09-18
Add server status endpoint

pierd:


app/src/server/mod.rs
```diff
@@ -125,20 +127,25 @@ pub async fn start(
     };
 
     // here the key is inserted into the asset cache
+    let server_state_holder = Arc::new(Mutex::new(None));
```
pierd, line 130: Any better ideas for this grappling hook?

philpax (approved):

PR #904, opened 2023-09-18
Remote shutdown and teardown of render loop

ten3roberts:


philpax (approved):

ten3roberts:
Are we using the dioxus example anymore?

ten3roberts:
If not, it may be a good call to remove it and avoid pulling in and building the whole tree for dioxus in the ci, as well as having a workspace and the double layers of toml files when editing dependencies for the web client

PR #915, opened 2023-09-19
Reduce logs on the client unless specified due to a performance and memory bottleneck

ten3roberts:


philpax:
I can confirm this helps, but it's still generating lots of unseen verbose logs - this is after about a minute: 
image
and the kinds of logs that are hidden are largely noise:
image

I think we need to go over our debug logs and remove some of them.

ten3roberts:
How did you test this?

When running locally the Dehug logs are enabled but comfigurable in the `main.ts` without requiring a rebuild

philpax:
Oh, I didn't realise. Are the debug logs disabled in production?

ten3roberts:
Yes.

Only `Info` and above are visible, since from what I've understood the Info log level is considered to be user facing, and it is good to give the users *some* idea of what is going on, rather than only warnings and such.

Many other rust projects default to having the info level user facing.

That being said, there is still nothing that prevents us to loop-print on the info level, but that is a bug on its own in such case and we should remove or downgrade the log in such case, just like with a warning level. I.e; logs are a tool, so they need to be used correctly and not abuse log levels (which we are working on).

Client script output is printed on the `Info` level, I think it is good to show the script output for the users; let me know what you think :)

ten3roberts:
We can at any time disable all logs or only filter warning levels on the website at any time without rebuilding the client and publishing a new version

philpax:
Great, yeah! Cool - yes, I was mislead by still seeing the debug logs go up. Setting `logFilter: "info"` resolves it, and memory usage stays constant. Nice work!

PR #922, opened 2023-09-20
Update Ambient version in manifests

### LINKED ISSUE #887: Set `ambient_runtime` version in examples `ambient.toml`
We should update the `ambient_runtime` version, same as we're updating the deployment id, so that when people clone the repo and try to run the examples it tries to use the last functional nightly version of ambient for the example

philpax:
Closes #887.

This makes Campfire update each manifest with the specified `ambient_version`, and adds a flag `--update-package-ambient-versions-only` that _only_ does this update so that it can be used as part of `deploy-packages` to update all of the ambient_versions of the packages after the initial reversion.

@pierd I could use your help with `deploy-packages.yml` - I was thinking that I'd insert a call to update-version --update-package-ambient-versions-only between these two:

```sh
cargo cf release update-version ${version}
git commit -a -m "Revert version back to ${version}"
```

so that it would look like

```sh
cargo cf release update-version ${version}
cargo cf release update-version ${deployed_version} --update-package-ambient-versions-only
git commit -a -m "Revert version back to ${version}"
```

but it occurs to me that there isn't necessarily an easy way to get the deployed version at that point. Can you think of a simple way, or is it easier to invert the flag and have it update everything *but* the ambient versions?

### DIFF
--- a/.github/workflows/deploy-packages.yml
+++ b/.github/workflows/deploy-packages.yml
@@ -74,7 +74,9 @@ jobs:
           # switch back to the main branch version
           version_majminpat=$(echo ${tag} | sed -E -e 's/^v([0-9]+\.[0-9]+\.[0-9]+).*$/\1/')
           version=${version_majminpat}-dev
-          cargo cf release update-version ${version}
+          # We do not update the Ambient versions of the packages to ensure that the nightly that they were deployed with
+          # remains active + accurate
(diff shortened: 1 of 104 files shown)


pierd:
IMHO inverting the flag might be easier (`--skip-packages`?). We also do a build in that operation so one call is preferred (of course we can address that separately).

Issue #909, opened 2023-09-19
`SIGHUP` each time you're exiting ambient


FredrikNoren:
I get this when I try to exit ambient (cmd-q):

![Screenshot 2023-09-19 at 12 12 57](https://github.com/AmbientRun/Ambient/assets/1288009/20bc95d3-3c53-403f-a132-b87f392cf543)

This is the callstack:

![Screenshot 2023-09-19 at 12 14 32](https://github.com/AmbientRun/Ambient/assets/1288009/5aacc001-8c91-4049-90b3-4f46fea99dfd)

FredrikNoren:
Seems like it might be related to the cli; the stack trace shown is in the cli.

Pombal:
SIGnal Hang UP is sent to the process when the controlling terminal is closed. It would be worth understanding why the cli breaks on it under a debugger and if the same happens with Ambient. Perhaps the cli needs to handle that signal with sigaction (SIGHUP, &sigaction, NULL) ...

FredrikNoren:
@ten3roberts Not sure. Can you try it on your mac?

FredrikNoren:
Ok this seems to be related to sentry; I just tried running it with a local build that doesn't have sentry and that doesn't trigger the breakpoint.

FredrikNoren:
Disabling sentry through the `settings.toml` file also fixes this.

FredrikNoren:
Add a line to the docs so people can find this issue here: https://github.com/AmbientRun/Ambient/blob/main/docs/src/user/setting_up_ide.md?plain=1#L14

PR #931, opened 2023-09-21
Randomly generated IDs

### LINKED ISSUE #679: Switch to randomly-generated IDs for packages
At present, package IDs are snake-cased IDs. However, we don't want to inherit the problems of a global namespace; instead, we've discussed this internally and we think the best way forward is to use a random ID with associated metadata (name, description, authors), and then offering additional searchability through the website/tooling.

That is, all existing package should go from `id = ambient_example_skinmesh` to `id = u3nu34tnu4rimsd` (RNG), and everywhere where the ID is used, a human-friendly alternative (name or import identifier) should be used instead.

### LINKED ISSUE #921: Remove `ambient view`
Don't think it works any longer and should anyway be part of the future IDE

philpax:
Closes #679, closes #921.

The code for the CLI was rearranged because my initial plan was to move all of the package commands into a `package` subcommand, and then set up top-level aliases, but I ran into a few hurdles and decided to leave it as-is for now.

I won't automerge this until I have signoff from the reviewers because I suspect we'll want to wipe the database before we do.


app/src/cli/package/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    pub fn package_path(&self) -> anyhow::Result<PackagePath> {
+        self.path.clone().try_into()
+    }
+}
+
+#[derive(Args, Clone, Debug)]
+pub struct HostCli {
```
pierd, line 84: I'm assuming this one is left here because of the hurdles.

app/src/cli/package/run.rs
```diff
@@ -0,0 +1,40 @@
+use ambient_native_std::asset_cache::AssetCache;
+use clap::Parser;
+
+use crate::{cli::ClientCli, client};
+
+use super::{serve, HostCli, PackageArgs};
+
+#[derive(Parser, Clone, Debug)]
+/// Builds and runs the package locally
+pub struct Run {
```
pierd, line 10: So the idea is for `run` to be under `package`? `ambient package run`?

shared_crates/package/src/manifest.rs
```diff
@@ -56,47 +56,102 @@ impl Manifest {
...
-        output[0] = *ALPHABET.choose(&mut rng).unwrap();
-        for i in 1..output.len() {
-            let idx = rng.gen_range(0..(NUMERALS.len() + ALPHABET.len()));
-            output[i] = if idx < NUMERALS.len() {
-                NUMERALS[idx]
-            } else {
-                ALPHABET[idx - NUMERALS.len()]
-            };
+        // TODO: see if there's a more intelligent way to ensure the first character
+        // is always alphabetic
```
pierd, line 89: Almost like BTC mining 😝

pierd (approved):

app/src/cli/package/mod.rs
philpax: Yeah, that's just the common args for the commands that host (run, serve). Can't do much about that, I think.

app/src/cli/package/run.rs
philpax: The idea was that I'd move all the commands there, and then have aliases back to them, as-is the case with campfire - so that you could always have a consistent interface, but also a convenient one. However, didn't happen, so it's still just `ambient run`.

shared_crates/package/src/manifest.rs
philpax: The comparison was not lost on me and it makes me sad 😂

FredrikNoren (approved):

philpax:
Waiting for #961 to go in, then merging

Pombal:
Worth retesting https://github.com/AmbientRun/Ambient/issues/909 if the CLI code changed.

philpax:
I think that issue is with the version-manager-CLI, not the Ambient CLI (yes this is confusing, see https://github.com/AmbientRun/AmbientCli/issues/3)

PR #940, opened 2023-09-22
Fix web client local dev

philpax:
Allows version mismatches on `www` and adds a `context` parameter to allow selecting another server

### DIFF
--- a/web/www/src/main.ts
+++ b/web/www/src/main.ts
@@ -82,7 +82,12 @@ import("ambient_web")
             return;
         }
 
-        ambient.init({ enableLogging: true, enablePanicHook : true, logFilter: "debug" });
+        ambient.init({ 
+            enableLogging: true, 
+            enablePanicHook: true, 
+            logFilter: "debug",
+            allowVersionMismatch: true,
+        });
 
         let target = window.document.getElementById("instance-container");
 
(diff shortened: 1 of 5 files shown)


web/www/src/main.ts
```diff
@@ -82,7 +82,12 @@ import("ambient_web")
             return;
         }
 
-        ambient.init({ enableLogging: true, enablePanicHook : true, logFilter: "debug" });
+        ambient.init({ 
+            enableLogging: true, 
+            enablePanicHook: true, 
+            logFilter: "debug",
+            allowVersionMismatch: true,
```
ten3roberts, line 89: ❤️

web/www/src/main.ts
```diff
@@ -92,8 +97,18 @@ import("ambient_web")
...
+
+        let params = new URLSearchParams();
+        if (packageId) {
+            params.set('package_id', packageId);
+        }
+        if (context) {
+            params.set('context', context);
+        }
+
+        const url = params.size != 0 && `https://api.ambient.run/servers/ensure-running?${params.toString()}` || "https://127.0.0.1:9000";
```
ten3roberts, line 111: I suspect this just allows joining a different hosted instance?

PR #943, opened 2023-09-22
Fix sky on the web

ten3roberts:
Fixes:
  - Tint and Wgpu/Naga disagreement of `textureSampleLevel` signature
  - Incorrect mie scattering, leading to a fuzzy/hazy sky
  - Unaccounted sun color for atmosphere, this is configurable by users, and should as such be reflected
  - Atmosphere remains if sun is dimmed
  - Add atmospheric absorption, the air isn't perfectly transparent
  - Fix circular cutoff in the void
  - Soften the small white disk that is the sun (or rather a plate in the sky)


ten3roberts:
This also makes the sun example start during the day and move slightly slower to be easier to see

philpax (approved):
Seems pretty reasonable, I trust that it works 🤣

ten3roberts:
Works on windows too 🙌

PR #945, opened 2023-09-23
bump as-any; get rid of downcast-rs usage

fogti:
NOTE: as the new `as-any` version implements `Downcast` directly for everything that could support it, name resolution changes a bit and requires more explicit dereferences.

* `cargo fmt --check` passes
* `cargo test --all` passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -556,13 +556,13 @@ dependencies = [
  "ambient_shared_types",
  "ambient_sys",
  "anyhow",
+ "as-any",
  "atomic_refcell",
  "bincode",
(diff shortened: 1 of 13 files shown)


FredrikNoren:
Hm the golden image test is failing because secrets aren't provided to external PRs. I'll just remove unity from the tests for now.

ten3roberts (approved):
Thanks, looks great.

Will trigger a rerun of the tests to see if golden images pass :)

FredrikNoren:
Seem like the GI was still running on old things so I'll just merge, this should be fine

PR #961, opened 2023-09-25
Fix pbr rendering

ten3roberts:
See commits messages for details

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1557,6 +1557,7 @@ dependencies = [
  "anyhow",
  "glam 0.24.1",
  "itertools",
+ "tracing",
  "yaml-rust-davvid",
 ]
 
--- a/crates/build/src/pipelines/materials/mod.rs
+++ b/crates/build/src/pipelines/materials/mod.rs
@@ -214,8 +214,8 @@ pub async fn to_mat(
         transparent: pipeline.transparent,
         alpha_cutoff: pipeline.alpha_cutoff,
         double_sided: pipeline.double_sided,
-        metallic: pipeline.metallic.unwrap_or(1.),
-        roughness: pipeline.roughness.unwrap_or(1.),
+        metallic_factor: pipeline.metallic_factor.unwrap_or(1.0),
+        roughness_factor: pipeline.roughness_factor.unwrap_or(1.0),
         sampler: pipeline.sampler.map(|sampler| SamplerKey {
             address_mode_u: sampler.address_mode_u,
             address_mode_v: sampler.address_mode_v,
--- a/crates/build/src/pipelines/models/quixel.rs
+++ b/crates/build/src/pipelines/models/quixel.rs
@@ -181,10 +181,9 @@ pub async fn object_pipelines_from_quixel_json(
                 metallic_roughness: Some(
                     pipe_image(&format!("{}_Roughness.jpg", quixel_id.resolution)).await?,
                 ),
-                roughness: 1.0,
-                metallic: 0.2,
                 ..Default::default()
             };
+
             let material_override = ModelImportTransform::OverrideMaterial {
                 filter: MaterialFilter::All,
                 material: Box::new(material.relative_path_from(out_materials_url)),
@@ -368,8 +367,6 @@ pub async fn object_pipelines_from_quixel_json(
                         .or_else(|| get_map_v2("components", "Roughness")),
                 )
                 .await?,
-                metallic: 0.2,
-                roughness: 1.0,
                 ..Default::default()
             }
             .relative_path_from(out_materials_url);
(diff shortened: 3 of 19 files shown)


guest/rust/examples/assets/unity/screenshot.png
ten3roberts, line 1: Could never get the unity example to work 😢

guest/rust/examples/basics/multiplayer/screenshot.png
ten3roberts, line 1: Look ma, it is finally yellow as supposed to 😄

guest/rust/examples/intermediate/dependencies/screenshot.png
ten3roberts, line 1: Hey look, the shininess is back just like in blender

ten3roberts:
@philpax after bumping the timeout for the CI the blessed images look different, as they are captured at 60s instead of 30s

The checking worked as it just tried every second until it mathed the 30th second or eariler if it is repeating at an interval, like the sun

guest/rust/packages/games/tangent/screenshot.png
philpax, line 1: Is the ground meant to be black here?

philpax:
Can you try removing the *2.2 constants in the following files:
https://github.com/AmbientRun/Ambient/blob/7aba88cca2eef35643e4a8c4f8657201afc317f1/guest/rust/packages/games/minigolf/src/server.rs#L160
https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/music_sequencer/src/client.rs#L124-L131
https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/tangent/mods/level_cubicide/src/server.rs#L124-L126

and see if they look as bright as they should? (The music sequencer I'm less sure about - from Arkanoid it seems like the colours are still dimmer than they should be for UI, but I'm guessing that's an entirely separate issue)

philpax (changes requested):
Awesome work! Glad to see this finally resolved 🎉 

Just wondering about a few things - if you can check those out and confirm how they should look/work, this is good to go.

guest/rust/examples/rendering/sun/screenshot.png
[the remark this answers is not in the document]
ten3roberts: Sortof, the screenshot is captured at 60 seconds instead of 30 seconds, as we bumped the golden images timeout

It has still passed, because the checking tries every second for a match

guest/rust/examples/assets/unity/screenshot.png
FredrikNoren: What's the problem with it?

ten3roberts:
> Can you try removing the *2.2 constants in the following files:
> 
> https://github.com/AmbientRun/Ambient/blob/7aba88cca2eef35643e4a8c4f8657201afc317f1/guest/rust/packages/games/minigolf/src/server.rs#L160
> 
> https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/music_sequencer/src/client.rs#L124-L131
> 
> https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/tangent/mods/level_cubicide/src/server.rs#L124-L126
> 
> and see if they look as bright as they should? (The music sequencer I'm less sure about - from Arkanoid it seems like the colours are still dimmer than they should be for UI, but I'm guessing that's an entirely separate issue)

How did we originally come up with the `2.2` value? Didn't know about this workaround 😅, but have always noticed the slight oddity of the color accuracy

FredrikNoren:
It feels like we should add a PBR example (the standard one, with just all the spheres) so we know that it works as intended. I added one here a while ago but it's on the host side: https://github.com/AmbientRun/Ambient/blob/main/crates/renderer/examples/pbr.rs

philpax:
> > Can you try removing the *2.2 constants in the following files:
> > https://github.com/AmbientRun/Ambient/blob/7aba88cca2eef35643e4a8c4f8657201afc317f1/guest/rust/packages/games/minigolf/src/server.rs#L160
> > 
> > https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/music_sequencer/src/client.rs#L124-L131
> > 
> > https://github.com/AmbientRun/Ambient/blob/6cb67f04bb6b132a8792139f756071c928ea1e71/guest/rust/packages/games/tangent/mods/level_cubicide/src/server.rs#L124-L126
> > 
> > and see if they look as bright as they should? (The music sequencer I'm less sure about - from Arkanoid it seems like the colours are still dimmer than they should be for UI, but I'm guessing that's an entirely separate issue)
> 
> How did we originally come up with the `2.2` value? Didn't know about this workaround 😅, but have always noticed the slight oddity of the color accuracy, it _does_ look like the default gamma, but in that case it is just a happy coincidence.
> 
> Just glad to have this fixed finally 🙌

Magic bodge constant. It's just a coincidence that it's the gamma correction factor, yeah.

philpax (approved):

philpax:
I'm just going to merge this. The issue is that the animation for `third_person_camera` changes each time the GI tests are run, because they all start at a slightly different time. We can figure out a better solution for this at another time (likely some kind of clock-override solution?)

Issue #951, opened 2023-09-25
Can't proceed past starting screen for Tangent or AFPS on the web


philpax:
Tangent and AFPS on the web send a reliable network message to the server to set their state (class and name respectively), but this doesn't seem to be processed by the server.

This suggests that
1) the client is not sending a message to the server correctly
2) the server is not handling the message from the client correctly
3) something else is happening

In any case, more investigation is required.

philpax:
Initial testing from @ten3roberts suggests that the WebTransport layer is working correctly (i.e. you can send unistreams from the client to the server). This suggests it's an issue with Ambient specifically.

TobiasEnholmX:
This does not happen everytime for tangent, but it does for AFP. For AFPS i can't get past putting in my name.

philpax:
I can't reproduce this on Tangent with:
- native client on local server
- web client on local server
- web client on remote server
- website + web client

@TobiasEnholmX How often are you seeing this happen?

Pombal:
Tangent seems OK but AFPS still gets an error

![image](https://github.com/AmbientRun/Ambient/assets/526714/d19b0c2e-b1a2-446b-badc-41119b4f97e3)

Pombal:
Now it worked so that's likely

PR #968, opened 2023-09-26
fix(input): use KeyboardInput to unify flow

philpax:
One half of #951. Looking into the other half.

This PR ignores control characters in `WindowKeyboardCharacter`, so that `TextEditor` will always use `WindowKeyboardInput` for its control behaviour.


philpax:
Tested this with macOS, macOS web, and Windows. Seems to work.

ten3roberts (approved):
Much better solution than the escape char workaround ❤️

PR #980, opened 2023-09-27
Tangent 0.2

philpax:
Tangent 0.2!

Now with a third-person character who can get in and out of spawned vehicles, plus some new mods:

- two choices of HUD
- character classes
- firerain environmental mod

Also, in the process of doing this, I discovered that the vehicles are approximately a third of the size they should be. Had to spend an hour rebalancing all of the stats...

Plus some minor fixes:
- `use_entity_{component,concept}` now get fired when the entity despawns
- `mod_manager` now includes local packages too


philpax:
Merging in because I didn't change anything with the GI tests/new package

PR #981, opened 2023-09-27
Clean up deploy output

FredrikNoren:
Just show the url that the user can click

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,7 @@ dependencies = [
  "anyhow",
  "axum",
  "clap 4.4.2",
+ "colored",
  "convert_case 0.6.0",
  "env_logger 0.10.0",
  "glam 0.24.1",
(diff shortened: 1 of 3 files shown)


philpax:
Not opposed in theory, but can we show the server URL on the website (not necessarily the deploy output) so that someone who can't use the web client can join natively?

i.e. I'm making a game and want to share it with my Linux-using friend who has Ambient installed (and our current target audience has a not-insignificant Linux userbase) - how do they join my deployed game?

FredrikNoren:
@philpax To share with friends that can't use the web client I would still recommend people to use the "download native" button on the website. It's the only reliable way to share it right now. Using `ambient join` if you have it installed won't work reliably since the version may mismatch.

philpax:
> @philpax To share with friends that can't use the web client I would still recommend people to use the "download native" button on the website. It's the only reliable way to share it right now. Using `ambient join` if you have it installed won't work reliably since the version may mismatch.

But that won't necessarily join them to the same server, and we're likely to have a userbase of dev-users who already have Ambient 0.3 launched - that is, people sharing their games with each other.

Joining with the wrong version will fail, so I think it's safe to expose that, even with that risk in mind.

FredrikNoren:
@philpax Alright

PR #982, opened 2023-09-27
Tangent reskin classes

philpax:
A mod for Tangent that, when loaded, reskins all of the classes, including existing ones, with Muscle Animals (not included).

Keeping this around as reference and/or something we can merge in if we find CC0 character models.


philpax:
On main through different means now

Issue #927, opened 2023-09-20
New external component definitions do not get synced to the user


philpax:
When ECs are registered, any connected clients are *not* made aware of them. This means that you can't actually add new components from packages to a running system with peers; we hadn't noticed because we've been testing locally and/or not using packages that add new components.

This also contributes to a problem where the web client can connect to the server before the latter has finished adding all of its packages; because it's raced its way in, it won't know of any of the ECs to come, and will end up failing once it receives `WorldDiff`s with those missing components.

The fix for this should use `on_external_components_change` to detect the addition of new components, and then make sure those definitions are made available *before* WorldDiffs are processed. I suspect we can reuse the DiffSerializer logic to inject the new component definitions.

PR #987, opened 2023-09-28
Sync external components through diffs

pierd:
Fixes: https://github.com/AmbientRun/Ambient/issues/927


philpax (approved):
Awesome work! Thanks for taking care of this 🙏

Issue #994, opened 2023-09-28
`ambient_sys` has several warnings on native


philpax:
```
warning: unused import: `Waker`
 --> crates/sys/src/control.rs:7:18
  |
7 |     task::{Poll, Waker},
  |                  ^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `removed`
  --> crates/sys/src/time/mod.rs:83:13
   |
83 |         let removed = self.heap.lock().remove(&Entry { deadline, timer });
   |             ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_removed`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `label`
  --> crates/sys/src/time/mod.rs:84:13
   |
84 |         let label = unsafe { &*timer }.label;
   |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_label`

warning: function `schedule_wakeup` is never used
  --> crates/sys/src/./native/time.rs:97:8
   |
97 | pub fn schedule_wakeup<F: 'static + Send + FnOnce()>(dur: Duration, callback: F) {
   |        ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function `sleep_until` is never used
   --> crates/sys/src/./native/time.rs:109:8
    |
109 | pub fn sleep_until(deadline: Instant) -> tokio::time::Sleep {
    |        ^^^^^^^^^^^

warning: function `sleep` is never used
   --> crates/sys/src/./native/time.rs:114:8
    |
114 | pub fn sleep(duration: Duration) -> tokio::time::Sleep {
    |        ^^^^^

warning: function `sleep_label` is never used
   --> crates/sys/src/./native/time.rs:118:8
    |
118 | pub fn sleep_label(duration: Duration, _: &'static str) -> tokio::time::Sleep {
    |        ^^^^^^^^^^^

warning: struct `Interval` is never constructed
   --> crates/sys/src/./native/time.rs:122:12
    |
122 | pub struct Interval {
    |            ^^^^^^^^

warning: associated items `new`, `new_at`, `tick`, and `set_missed_tick_behavior` are never used
   --> crates/sys/src/./native/time.rs:127:12
    |
126 | impl Interval {
    | ------------- associated items in this implementation
127 |     pub fn new(period: Duration) -> Self {
    |            ^^^
...
[...]

PR #997, opened 2023-09-29
fix: warnings on native and web

ten3roberts:
Fixes: #994


ten3roberts:
This fixes warnings that arose from recent refactorings, mostly due to import only being for one of wasm or native, and being unused otherwise.

This also fixes new warnings that were introduces to recent version of Rust

PR #1017, opened 2023-10-02
Add simple frame dropping

pierd:


pierd:
Waiting for results of testing of https://github.com/AmbientRun/Ambient/pull/1018 before merging since we might not need frame dropping after all

ten3roberts (approved):

PR #1024, opened 2023-10-03
Fix perf2

ten3roberts:


ten3roberts:
> Instead of commenting out the code, could you chuck all of it under a `const DEBUGGER_ENABLED: bool = false; if DEBUGGER_ENABLED {}`, or failing that, conditional compilation?
> 
> Otherwise lgtm

Can I do that outside this PR on main, and we just cut this release now?

philpax:
I'm happy to push the release to get this in - I want to avoid as many discrepancies between main and the RC as possible.

Issue #1004, opened 2023-09-30
Outlines break the renderer on web


philpax:
Create a new project with `ambient new`.

Replace the bottom half of `server.rs` with
```rust
let quad_id = Entity::new()
    .with(translation(), vec3(0., 0., 0.))
    .with(quad(), ())
    .spawn();

println!("Hello, Ambient!");

fixed_rate_tick(Duration::from_secs(1), move |_| {
    if entity::has_component(quad_id, outline_recursive()) {
        entity::remove_component(quad_id, outline_recursive());
        entity::add_component(quad_id, color(), vec4(1.0, 1.0, 1.0, 1.0));
    } else {
        entity::add_component(quad_id, outline_recursive(), Vec4::ONE);
        entity::add_component(quad_id, color(), vec4(1.0, 0.0, 0.0, 1.0));
    }
});
```

Observe that this works as expected on native.

Next, serve the same package, but connect to it using the web client (`cargo cf web serve`).

Note that as soon as the outlines appear, the renderer freezes, and the console is spammed with the following:

```
Blending is enabled but color format (TextureFormat::RGBA32Float) is not blendable.
 - While validating targets[0].
 - While validating fragment state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor "standard_shader_flat_material_shader_true"]).

249[Invalid RenderPipeline "standard_shader_flat_material_shader_true"] is invalid.
 - While encoding [RenderPassEncoder "Outlines"].SetPipeline([Invalid RenderPipeline "standard_shader_flat_material_shader_true"]).

248[Invalid CommandBuffer "GameState.render" from CommandEncoder "GameState.render"] is invalid.
 - While calling [Queue].Submit([[Invalid CommandBuffer "GameState.render" from CommandEncoder "GameState.render"]])
```

The renderer recovers after the outline is removed.

ten3roberts:
Looks to be the same as #888

Should we add it back to the milestone, as it was removed recently?

philpax:
Nah, I just removed the code using the outlines for now. We can fix this later.

pierd:
I just hit this for Tic Tac Toe example.

![image](https://github.com/AmbientRun/Ambient/assets/752792/53a3ca5f-d857-468e-93d1-a5d2db41c12f)

![image](https://github.com/AmbientRun/Ambient/assets/752792/e772d5b8-4578-427f-9c25-04d5734033d9)

philpax:
Done

Issue #888, opened 2023-09-18
Tictactoe bug: Color blending is broken, uses non-blendable color format


ten3roberts:
Tictactoe is not working: https://ambient-733e7.web.app/packages/ambient_example_tictactoe

philpax:
I've temporarily disabled the outlines for tictactoe and editor - should be easy enough to turn them back on when this bug is fixed

PR #1027, opened 2023-10-03
Fix outline rendering

ten3roberts:
This was caused by `RgbaF32` not being blendable according to the spec, but native vulkan being lenient

Fixes: #1004 Fixes: #888


philpax:
Great work!

PR #1039, opened 2023-10-04
Character animation performance fix + benchmark

FredrikNoren:
We were setting strings constantly which has a huge cost across the guest/host boundary currently.

### DIFF
--- a/guest/rust/Cargo.lock
+++ b/guest/rust/Cargo.lock
@@ -688,6 +688,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "animations"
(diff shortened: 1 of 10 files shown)


ten3roberts (approved):

Pombal:
@FredrikNoren Are you sure it's working? The new version only contains T pose, it's not running animations
https://ambient.run/packages/tx4tcaxvwbfmaszwtndpovrmdysu5wvr

FredrikNoren:
Yeah I'm not sure of anything right now. Closing

Issue #1038, opened 2023-10-04
Linux WebGPU initialization failure is not propagated upwards


philpax:
Linux Chrome doesn't support WebGPU in totality yet, but still registers a `gpu` object, which means the runtime attempts to initialize. Unfortunately, this fails and the error is not propagated upwards, so it is indefinitely "Loading" with the only sign of trouble in the web console.

We should do one or both of these things:
- [x] change web embed to check for Linux and fail fast before initialization
- [x] propagate errors in initialization to the web embed, so users know that there are failures in that step

ten3roberts:
Error are propagated from `AppBuilder::build` to the instance manager.

I did that in the merge of `init` and `start` last week, where the whole `start` function returns an error if `AppBuilder::build` failed, and that works properly.

The issue is that the Gpu initialization just unwraps in the asset cache and doesn't return an error at all.

Fixing it

PR #1043, opened 2023-10-05
Propagate gpu creation errors

ten3roberts:
Fixes: #1038


philpax (changes requested):
Error handling logic looks good, but can report the error in a user-facing way? (You might be able to attach a ‘<p>’ or such). That way, the user will be made aware of the error without having to check the console, and they should be able to copy it out.

ten3roberts:
image

This will then show up on the website as an error screen, because the website awaits that result

philpax:
Oh, right - I see, the website handles that differently, it's just `www` that logs. Got it.

philpax (approved):

PR #1044, opened 2023-10-05
fix(web): water using the same trick as the atmosphere

ten3roberts:
The textureSampleLevel is buggy on wgpu and does not accept a int for depth comparison.

Is there a specific reason the compare*Level* was used, rather than a normal texture sample. Did it fix a specific bug in the past? The git log goes back to the `Init` squash.


ten3roberts:
image

PR #1061, opened 2023-10-09
Use --feature production on cf deploy-all

FredrikNoren:
The `production` feature flag is needed to enable `assimp`, which is needed for to deploy the assimp example. It's only adding that feature flag for deploy-all now though.

It feels like deploy-all should be moved to `ambient` runtime instead though, that that the ci job should just use the latest ambient runtime? (Instead of building at again)

### DIFF
--- a/campfire/src/join.rs
+++ b/campfire/src/join.rs
@@ -58,5 +58,5 @@ pub fn main(join: &Join) -> anyhow::Result<()> {
         args.extend(join.params.args.iter().map(|s| s.as_str()));
     }
 
-    run_ambient(&args, join.params.release)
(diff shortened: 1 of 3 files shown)


pierd:
I was planning on making it so we don't have to rebuild ambient to deploy. Yeah, we should be able to use ambientcli.

FredrikNoren:
@pierd Don't know if we _need_ to use ambientcli, could also just download artifacts from previous builds (like the golden images do; they download the ambient runtime executable).

pierd:
Created an issue: https://github.com/AmbientRun/Ambient/issues/1069

Issue #1050, opened 2023-10-06
`ambient_git_rev` is triggering a rebuild of everything when you commit


FredrikNoren:
We might need to reverse things and put a `static GIT_REV` somewhere, and only run the git operations at the very top (in the `app`).

PR #1062, opened 2023-10-09
Fix rebuild on commit

pierd:
https://github.com/AmbientRun/Ambient/issues/1050

With this a new commit makes cargo rebuild only `ambient_git_rev_init` and `ambient` crates.


ten3roberts:
How does this function on the web client?

pierd:
> How does this function on the web client?

Good point 😅

ten3roberts:
> > How does this function on the web client?
> 
> Good point 😅 It most likely doesn't. I wonder if our CI will catch this.

Most likely it'll pass.

The CI at the moment cathes many things, mostly because of denied types and methods, and a quite extensive abstraction layout in use, which has really proven to squash down exponential complexity.

It does however not catch differences between `app` and `web/client`, so that is the golden rule of thumb, if you change a crate, all is likely fine. If you change the init code, make sure to have a glance at the web client

PR #1065, opened 2023-10-09
Allow for Chrome app selection in `campfire web open-browser`

pierd:
I just want to be able to run: `cargo cf web open-browser -a 'Google Chrome Canary'` on my mac. Not sure how much it makes sense for other OSes. Tested on:
- [x] macOS
- [x] Windows
- [ ] ~~Linux~~ - I don't have Linux installed anywhere that is useable for this.

Also fixes URL passing on macOS.

### DIFF
--- a/campfire/src/web/browser.rs
+++ b/campfire/src/web/browser.rs
@@ -1,10 +1,17 @@
 use anyhow::Context;
+use clap::Args;
 use sha2::{Digest, Sha256};
 use spki::EncodePublicKey;
 use std::process::Command;
 use x509_certificate::X509Certificate;
 
-pub async fn open() -> anyhow::Result<()> {
+#[derive(Debug, Args, Clone)]
+pub struct BrowserOptions {
+    #[arg(short, long)]
+    pub app: Option<String>,
+}
+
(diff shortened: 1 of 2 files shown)


campfire/src/web/browser.rs
```diff
@@ -74,10 +81,16 @@ fn spawn(spki: &str, _url: &str) -> anyhow::Result<()> {
 }
 
 #[cfg(target_os = "linux")]
-fn spawn(spki: &str, url: &str) -> anyhow::Result<()> {
+fn spawn(spki: &str, url: &str, options: BrowserOptions) -> anyhow::Result<()> {
     const CANDIDATES: &[&str] = &["google-chrome", "google-chrome-stable", "chromium"];
+    let provided_candidate = &[options.app.as_deref().unwrap_or("google-chrome")];
+    let candidates = if options.app.is_some() {
```
ten3roberts, line 87: 👍 Good strategy to disable fallbacks if something explicit is passed, and not the other way around.

Have fallen victim of "you think you specified an override, but it was incorrect so a fallback was silently used for 3 months" *many* times 😅

ten3roberts (approved):

Issue #1033, opened 2023-10-04
Cargo.toml ambient_api 0.3.0 still points to git repo


FredrikNoren:
Repro:

1. `ambient new hello`

In Cargo.toml it says: `ambient_api = { git = "https://github.com/AmbientRun/Ambient.git", tag = "v0.3.0" }`, should be `ambient_api = "0.3.0"`

PR #1070, opened 2023-10-10
Fix `ambient_api` dep version in new project

pierd:
https://github.com/AmbientRun/Ambient/issues/1033

Not sure if we should include release candidates in this too - waiting for opinions.

Tested with `cargo cf release update-version 0.3.1` followed by `cargo run -- new ...`.


FredrikNoren (approved):
Hm might be nice with RC's but doesn't really matter, this covers the most important part.

Issue #584, opened 2023-07-18
Address misuse of log and tracing verbosity levels


ten3roberts:
We currently mainly use the `info` levels and above for our logging directives. This has resulted in a lot of verbose terminal output, which we have fixed by conditionally reducing log levels by injecting extra directives.

This is not the intended usage of logs and env filtering. The extra steps we take to crate-by-crate silence the info log level is very rough and leads to a *very* inconsistent behavior, and makes it difficult to conditionally enable some extra logging for a crate you are trying to debug. It is not possible because there is no way to distinguish the lower level chatty logs from the general high level logs since they are all *log::info*.

Setting a specific log level using `RUST_LOG` often does not lead to the intended result, and more often than not leads to a complete flood of logs since we don't distinguish them.

The solution to this is *not* to add more fine grained *silencing* but rather make proper use of the log levels. This way, we can also add *debug* or even *trace* level logs that are very chatty but can be kept in the source and enabled explicitly (rather than *unsilence* through a combination of a complicated directives in an env string"). If a bug is encountered in a crate the crate or even module can be surgically targeted for `trace` directives to be enabled, without leading to unexpected walls of texts from other logs that surface.

Additionally, the current method makes logging inconsistent when a function crosses a crate boundary via a helper function and your logs suddenly dispensary.

What do you think about this @philpax

ten3roberts:
I may add that on the web, there is no env filtering, since there is no environment variables.

This means that the only way to filter logs is through chrome's `verbosity` checkbox in the console.

Our current situation means it is a boolean of *all logs* or *nothing*, and makes it incredibly unwieldy to use.

philpax:
I think this might be a dupe of #459. When you have some free time, would you be interested in fixing this up and closing both issues?

philpax:
I'd say they're both addressing the user problem of "there are too many logs", but I'm not too fussed. We can address both at the same time. (Perhaps #459 is better linked to #564)?

ten3roberts:
Yes we could. #584 is a *solution* to the end-user problem defined by #459, while #564 aids our development experience after #584

Issue #459, opened 2023-05-31
Hide logging output by default


FredrikNoren:
It's getting really noisy when starting ambient now. I think it's time that we add a `--verbose` or `--logging` flag, and only output what a user prints with `println` from with the scripts by default + "use `ambient join eu.proxy.ambient.run:9165` to join". We could also utilize temporary console lines while the project is building (i.e., showing "Building asset/..." while it's building, but then removing that line).

ten3roberts:
I am quite familiar with how the logging setup works at the moment, so I could see myself doing this.

Additionally, I think that we shoud remove the fallback log directives which silent some specific crates, so that `RUST_LOG=info` does show infos for e.g `ambient_gpu`. This behavior has shot us in the foot a few times and is quite a surprising behavior as you get the false premise that you now have all logs of info level.

I think removing this and having a `--verbose` flags will increase consistency and make the overall user experience better.

Furthemore, there are some amount of logs in the server modules, and they will be addressed in the #421 PR

philpax:
100% agreed with this. I cleaned up the logs earlier, but INFO logs keep finding their way into the codebase. I'm fine with removing the per crate silencing if we're much more judicious about what we log by default.

ten3roberts:
Should we use tracing and be more judicious with our log levels, like using `debug` for most spammy stuff rather than doing `info` for everything, and then selectively silencing `info`?

For user facing messages we could just use `eprintln` or `println`.

What do you think?

PR #1073, opened 2023-10-11
Fix misuse of overly verbose log levels and unify use of log and tracing

ten3roberts:
This uses the appropriate log levels for our logging. 

The recommendation is:
`info` is surfaced to, and used as a means to communicate with the users, `debug` is intended for lower priority and implementation oriented messages, and trace is for very verbose and detailed information, most commonly includes debugged structs, or very implementation specific information that is almost entirely useful for internal devs and never users.

Additionally, this means the prefixed log filtering that were a common source of surprising behavior has been almost entirely removed. The only remaining targets are external crates that don't use logging appropriately.

The split usage of log, and its superset replacement *tracing* has been fixed, and tracing is used throughout. Tracing has lot of more useful features, and is especially good at capturing spatially temporal information through the use of spans. 

Tracing also has support for further detailed and structured logging, async, and much more, and is what is the new standard in the wider rust ecosystem.

Fixes: #584 
Fixes: #459

### DIFF
--- a/app/src/client/wasm.rs
+++ b/app/src/client/wasm.rs
@@ -22,7 +22,7 @@ pub fn initialize(
                 MessageType::Info => tracing::info!(%module_name, "{}", message),
                 MessageType::Warn => tracing::warn!(%module_name, "{}", message),
                 MessageType::Error => tracing::error!(%module_name, "{}", message),
-                MessageType::Stdout => tracing::info!(%module_name, "stdout {}", message),
+                MessageType::Stdout => tracing::info!(%module_name, "stdout: {}", message),
                 MessageType::Stderr => tracing::info!(%module_name, "stderr: {}", message),
             };
         },
--- a/app/src/client/wasm.rs
+++ b/app/src/client/wasm.rs
@@ -15,21 +15,16 @@ pub fn initialize(
     mixer: Option<AudioMixer>,
 ) -> anyhow::Result<()> {
     let messenger = Arc::new(
-        |world: &World, id: EntityId, type_: MessageType, message: &str| {
-            let name = world.get_cloned(id, module_name()).unwrap_or_default();
-            let (prefix, level) = match type_ {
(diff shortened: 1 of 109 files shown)


ten3roberts:
# Old

image

**Note**: this uses our meticulously crafted specifc filtering rules that makes *enabling* additional logs very inconvenient as you can't differentiate between logs

# New

image

app/src/client/wasm.rs
```diff
@@ -15,21 +15,16 @@ pub fn initialize(
...
-                level,
-                "[{name}] {prefix}: {}",
-                message.strip_suffix('\n').unwrap_or(message)
-            );
+            match ty {
+                MessageType::Info => tracing::info!(%module_name, "{}", message),
+                MessageType::Warn => tracing::warn!(%module_name, "{}", message),
+                MessageType::Error => tracing::error!(%module_name, "{}", message),
+                MessageType::Stdout => tracing::info!(%module_name, "stdout {}", message),
+                MessageType::Stderr => tracing::info!(%module_name, "stderr: {}", message),
```
pierd, line 26: Any reason why stderr has `:` and stdout doesn't?

pierd (approved):

app/src/client/wasm.rs
pierd: Also maybe let's use error level for stderr

ten3roberts:
Update: 

A slightly more compact format

image

Additionally, there is a feature flag `tracing-tree` `cargo run --features tracing-tree -- run ...` that makes it *much* clearer which logs come from which scope and how they are grouped both spatially and temporally.

image

ten3roberts:
I mean, just *look* at this 😁 

image

This is so useful for debugging contexts and knowing how logs related to each other

ten3roberts:
This `<unknown time>` is fixed as well

image

PR #1078, opened 2023-10-12
Remove temporary sleep for server startup

ten3roberts:
@pierd do we still need this soft server start, or is probing and initialization order working correctly now?


pierd (approved):
It shouldn't be needed anymore.

PR #1085, opened 2023-10-13
Fix lints

pierd:


ten3roberts (approved):
I see, now you will be able to finally compile (somewhat) 😝

pierd:
> I see, now you will be able to finally compile (somewhat) 😝

Just removing the annoyances one by one ✂️

Issue #827, opened 2023-09-08
Potential memory leak when running instancing example


pierd:
Running instancing example in the cloud without any clients connected uses ~200MB RAM. One connected client bumps the memory usage to ~900MB, disconnecting doesn't bring that memory down. Connecting again increases the memory usage even more (eventually the server was killed).

![image](https://github.com/AmbientRun/Ambient/assets/752792/5943603b-bc86-41df-9d35-5a4f80d09506)

pierd:
<img width="512" alt="Screenshot 2023-09-08 at 19 29 50" src="https://github.com/AmbientRun/Ambient/assets/752792/50ae1bc0-1824-45ce-a5cf-fd4a8c1eab30">

Might be something in the networking code as I've noticed a similar thing in afps after unclean disconnect (client crash).

pierd:
Good news, the issue is with the diffs queue. When a player's bandwidth is not fast enough to receive all the updates, we queue them up while we send the previous one. That queue is held in raw format (not serialized and not even deduplicated or merged). With artificially limiting the bandwidth to 1MB/s (instancing requires about 7MB/s) I was able to reproduce having ~700MB worth of data in the queue for that player.

I'm planning to fix it by merging and serializing the diffs while the previous data is in flight.

ten3roberts:
It's always the queues :smh:

PR #1094, opened 2023-10-17
Make `handle_diffs` process diffs queue while sending

pierd:
To fix queue memory usage when clients can't receive diffs quickly enough: https://github.com/AmbientRun/Ambient/issues/827

Note: it only helps with `SetComponents` changes, shape changes are not compacted and will still occupy memory.


crates/network/src/proto/server.rs
```diff
@@ -378,55 +379,91 @@ pub async fn handle_diffs<S>(
...
+    let mut intermediate_diff = IntermediateWorldDiff::default();
+
+    let mut stream = Some(stream);
+    let mut sending_future = OptionFuture::default();
+
     loop {
-        let msg = tokio::select! {
+        tokio::select! {
+            // check if sending has finished (if any)
+            Some(sending_result) = &mut sending_future => {
```
ten3roberts, line 391: The future taketh, and the future giveth 😝

Issue #1109, opened 2023-10-23
Docker image cannot execute `ambient build` ?


kevzettler:
I'm trying to use the Docker image to build and host.

```
docker run -v /root/mygame:/data \
ghcr.io/ambientrun/ambient:sha-2c02f6c19c92c3751cbbde2a532870b5120b0018 \
/app/ambient build /data
```

This fails with:
```
15:00:54  INFO build_package: ambient_build: Building package "aceclone" (lsvr7wpee4biw3sleyscm7har4xzjvoa) name="aceclone"
15:00:54  INFO build_package: ambient_build: Assets built, building source code... name="aceclone"
Error: Failed to build Rust in "/data/build/lsvr7wpee4biw3sleyscm7har4xzjvoa"

Caused by:
    `rustup` is not installed. Please install it with https://rustup.rs/ for the best experience.
```

 The container doesn't include rustup? I assumed that the Docker container would create a complete `ambient` executable that could be used to both build and host a project. But the Dockerfile has comments  https://github.com/AmbientRun/Ambient/blob/b555277d01466a6034191fe982b9d4f72457e2ee/slim-release.Dockerfile#L1-L5 about running the build separately on the host machine and then using the container to serve. This seems unfortunately redundant because that would mean you need to install the ambient binary on the host machine and install the Docker image to use the container.

FredrikNoren:
@pierd can you take a look? Think you know this the best

pierd:
`slim-release.Dockerfile` is optimised for size (hence *slim*) and that's why it doesn't include rust toolchain.

We should be able to create one with fully functional Ambient setup that will allow for building too - I'll do that.

pierd:
@kevzettler The most recent nightly image (`ghcr.io/ambientrun/ambient:sha-f155798b404ecd3e02e36659f070e97422a816ed`) should allow for building.

Note: If you are using the Dockerfile directly then I unified the ones we had and not there's only one to rule them all.

PR #1116, opened 2023-10-25
Turn slim-release into a full release

pierd:
https://github.com/AmbientRun/Ambient/issues/1109

We currently extract the binary that is built in docker and upload it separately (pending switching to regular ambient builds) so we don't really need the build to be slim.


philpax (approved):
LGTM, but there might be some overlap with the `Dockerfile` at the root of the repository. We should probably remove that or clarify its relation.

PR #1123, opened 2023-10-27
Adjust frame dropping frame time calculation

pierd:
... so it calculates from the beginning of the last frame to the beg of the current one. It's more exact when processing out of our task takes significant time.


philpax (approved):

pierd:
I've discovered something new and I still need to do more testing.

It still doesn't work well on my mac's builtin display (120Hz). I'm going to make it smarter and detect the frame rate instead of having a hardcoded one.

PR #1126, opened 2023-10-29
fix cli http_interface_port consistency

kevzettler:
Seems this cli help text is incorrect and code actually defaults to `8999` at:

https://github.com/AmbientRun/Ambient/blob/256172a79d203f0669dce6bf5c9a7fab0cbab5d1/app/src/server/mod.rs#L303


pierd (approved):

pierd:
Nice catch, thanks.

PR #1140, opened 2023-11-05
Add hide component for nameplates

kevzettler:
This PR attempts to add a parameter to t he `nameplates` package that does not render the name plate for the local player. 

This doesn't currently work. I'm opening PR for discussion: 

In my game code I attempt to use with:
``` rust
    spawn_query((is_player(), user_id())).bind(move |players| {
        for (player_id, (_, uid)) in players {
            // First, we check if this player is the "local" player, and only then do we attach a camera
            if uid == entity::get_component(entity::resources(), local_user_id()).unwrap() {
                // hide the nameplate for local player
                println!("adding hide to {:?}", player_id);
                entity::add_component(player_id, packages::nameplates::components::hide(), ());
                p
                ...
```

However this doesn't work as the `nameplate` package is executed before this `hide` component gets added.

initially discussed at: https://discord.com/channels/894505972289134632/1141326292919263262/1170466566693986324

### DIFF
--- a/guest/rust/packages/std/nameplates/ambient.toml
+++ b/guest/rust/packages/std/nameplates/ambient.toml
@@ -18,4 +18,7 @@ text_size = { type = "F32", name = "Nameplate Text Size", description = "The tex
     "Debuggable",
     "Networked",
 ] }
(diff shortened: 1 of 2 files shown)


kevzettler:
This looks good and works @philpax thanks a lot, please merge!

Issue #1143, opened 2023-11-07
Make Ambient server support HTTPS for asset serving when certs are provided


pierd:
so that Ambient server can be use for hosting both the assets and [the web client](https://github.com/AmbientRun/Ambient/issues/1133) ([WebGPU requires Secure context](https://developer.mozilla.org/en-US/docs/Web/API/WebGPU_API)).

Issue #1133, opened 2023-11-01
Make Ambient server serve Ambient web client


pierd:
Currently running `ambient serve ...` runs the server that can be used by both native and web Ambient clients.

We should add an option for it to also host Ambient web client that automatically connects to itself for easier self-hosting and local testing. For example `ambient serve --host-web-client ...` would make Ambient server's HTTP server host web client code (with potentially additional options for port overriding and server hostname).

pierd:
Added simple hosting in https://github.com/AmbientRun/Ambient/pull/1153. I don't want to over-engineer port overriding and other features so I'm closing this and we can revisit if there's a need for more configuration options.

PR #1153, opened 2023-11-08
Make it easier to self host Ambient servers

pierd:
- `--use-https` options to serve http content with TLS so that it conforms to Secure context restrictions (https://github.com/AmbientRun/Ambient/issues/1143)
- simple embedded web client page hosting when running a released version (https://github.com/AmbientRun/Ambient/issues/1133)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,7 @@ dependencies = [
  "ambient_world_audio",
  "anyhow",
  "axum",
+ "axum-server",
  "clap 4.4.6",
  "colored",
  "convert_case 0.6.0",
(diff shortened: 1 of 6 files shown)


philpax (approved):
Awesome work, as always. Didn't occur to me that you could just use the prebuilt web client for this.

Only thing I'd suggest is having a fallback page for unreleased versions that's like `<h1>Unreleased versions do not support the self-hosted web client</h1>` to avoid confusion.

pierd:
FYI @kevzettler this should be available in the most recent nightly (0.3.2-nightly-2023-11-09)

Issue #1148, opened 2023-11-08
Package documentation generation


philpax:
At present, there's no easy way to see what items are available to your package or its dependents. We should implement a solution like `cargo doc` for this.

This will be a multi-step process:
1) https://github.com/AmbientRun/Ambient/issues/988
2) Have the Ambient build process generate a JSON with all of the items, similar to Rustdoc's JSON output
3) Build `ambient doc`, which consumes that JSON (and/or builds it if not yet available), and outputs a static site of docs
4) Update `ambient deploy` so that it will build the docs and upload them as well
5) Change the schema view on the website to show said generated docs (in an iframe or something, or just as a straight link)

I considered consuming that JSON with Dioxus and building a dynamic view, but a static site is easier to ship around and search with conventional browser tools, and doesn't require building and shipping the Dioxus website with Ambient itself.

philpax:
We might also want to include the built assets in the resulting JSON and show them within the doc'd website. Bonus points if we embed an asset viewer so that you can see what the built assets look like, but we might need a slim version of Ambient for that. (Or perhaps #1133 ?)

PR #1171, opened 2023-11-17
Package projection build script + static documentation generation

### LINKED ISSUE #988: Ambient package projection as a generated file, not as a macro
At present, we project your Ambient packages to the Rust code through a macro. However, this macro has proven to be quite troublesome (#907 #908 and several other issues resulting from "has this macro actually run?")

I'd like to propose decoupling the codegen from the `main` macro, so it can focus on adapting the entrypoint to teh component entrypoint, and having `ambient build`/`build.rs` generate a `generated.rs` for you.

Benefits:
- it will now be very obvious when the generated code updates, unlike the proc macro, which you can never be sure if it's run or not
- it will resolve all of the issues with the macro not working, or generation breaking in non-obvious ways
- you can see what's generated (!), which will help with discoverability, even if it's obviously macro-generated code (we can improve the quality of output)
- your rust-analyzer will no longer be held up on making network requests that may or may not resolve
- we can now safely include the built asset URLs within the output, as we can guarantee that the generated code is created after the assets are built, but before the Rust code gets built
- it aligns it with all of our other package-based codegen (the host and API both use a `generated.rs` solution for similar reasons)

Disadvantages:
- you'd have to run `ambient build`/another command to update the `generated.rs` after changing the manifest
- it's not as automatic/automagic as the current solution, but that might actually be a good thing (people have found it confusing to understand what's going on)
- it may not be immediately obvious to a new user why none of the imports exist - we could alleviate this by generating it automatically on new package, and/or including it in Git (unsure about this)

Side notes:
- it's possible that this codegen could be done in `build.rs`, as is the case with `ambient_wasm` and `ambient_ecs`. not sure if that would have any issues.
  - there are likely some nuances here around when-to-build, especially as your dependencies can change from underneath you 
- the future Ambient IDE could automatically update this generated code for you if you update the schema

philpax:
Closes #988 and #1148.

Removes the codegen from the Ambient `main` macro, making it *much* more responsive. In exchange, you'll need to update your `Cargo.toml`/modules and add a `build.rs`. See the CHANGELOG for more information.

The Ambient build process now produces an `ambient_package.json` which contains all of the items in the semantic system, similar to the `rustdoc` JSON.

This is then consumed by `ambient_package_docs` (may rename to `ambient_package_docgen`) to produce a fully-static website, similar to `rustdoc`, that lets you browse through all of the packages available to the package you're generating docs for.

image
image
image

The styling is temporary until we make some time to properly redesign it. Until now, it's functional-brutalist-programmer aesthetic! (Complete with semantic colouring.)

There are a few small things that don't quite work properly, a few missing conveniences (i.e. buttons to copy, etc) and documentation isn't shared between packages, so `ambient_core` has its documentation generated for every single package. We can tweak that in future; I just wanted to get a functional prototype out there.

Known issues:
- [x] It's not obvious how to actually get to your package's documentation. Could add an `ambient doc` that builds it and opens it with `--open`, or add `--open-docs` to `ambient build`.
- [x] There's no way to return to other packages once you get to the primitive types, which don't belong to any package. I'm thinking that these should probably just be moved into `ambient_core` anyway.
- [x] There's a *lot* of unnecessary copies when you're building a tree of packages (i.e. if you have packages A, B, C, D, each depending on the previous, then there will be 1+2+3+4 packages of docs generated). Maybe docgen only happens at the top level, or some kind of sharing happens?
- [x] Cross-package links don't contain the package name, so a concept using components from Ambient Core will not show "Ambient Core" in the path

Issues:
- [ ] Move `packages.rs` to OUT_DIR
- [ ] There are no buttons to copy anything. Not sure how that should work when the package ID will depend on the package you're using it from.
- [ ] Still needs to be integrated into the website. Given it's using regular page navigation, I suspect this will just be a link to the generated docs.
- [ ] It doesn't contain any information about the built assets, so asset-only packages look a bit barren. The hard bit is that we don't know about the build metadata for the packages that we aren't building - need to figure out how to get that information.
- [ ] Collate all `OutAssets` in build
- [ ] There's no way to search for items.
- [ ] Ideally, there's some form of cross-reference, so you can see what uses a given component / type / etc.


FredrikNoren:
@philpax Great stuff! So there's no way we can put the build.rs in ambient_api instead? (So people don't have to create it themselves)

philpax:
I tried a few different solutions to that, but as far as I can tell, Cargo doesn't expose where the "root" package is (i.e. if I'm building `tangent` which has a dependency on `ambient_api`, `ambient_api`'s build script doesn't get told where `tangent` is). 

I agree that having to create the `build.rs` isn't ideal, but it should hopefully be mitigated by being created by `ambient new`.

FredrikNoren:
@philpax Hm yeah especially in a workspace `ambient_api` would only be built once I assume :/ Maybe it could glob for all `ambient.toml` files or something, based on workspace root? Think you could get the root by walking backwards from `CARGO_TARGET_DIR` until you find a `Cargo.toml`, but not sure how stable that would be

philpax:
Yeah, that would make a lot of assumptions about the user's workspace / directory layout and seems like it could break easily. 

Looking at the env variables available to the build script for `ambient_api` (which only runs once, as you've noticed), there doesn't seem to be anything we could reliably use as a locator here.

I think we're stuck with this for now; I'll see if I can ask around the Rust community, but I'm pretty sure this level of sandboxing is intentional to prevent dependencies from messing up crates higher up in the hierarchy.

FredrikNoren:
@philpax Hm wait, does this output a rust file in the src directory? It should output one in the target dir right?

FredrikNoren:
Here's an example how it's done in the design tokens crate: https://github.com/AmbientRun/DesignTokens/tree/master/crate
In main.rs you'd just do `mod package { include!(concat!(env!("OUT_DIR"), "/ambient.rs")); }`

philpax:
Yep, you're right. I didn't do that because I thought the IntelliSense would be broken/flimsy (based on past experience), but it seems to work fine. I've made the change and purged all of the `packages.rs` - they'll be gone in the next push.

philpax:
@FredrikNoren Okay, damn, IntelliSense *is* a bit strange with that solution. It works fine in the workspace, but packages created outside of it don't have working IntelliSense on `packages`, even after reloading the window.

Generating into `src/` means the IntelliSense always works and is always up to date, and it's also obvious where the file's coming from.

Maybe it should just go into `src/` again, but `src/packages.rs` is in the default `.gitignore`?

FredrikNoren:
Ah, that sucks :/ Hm yeah guess .gitignore would help at least. Could also post an issue on rust-analyzer (it should work right?)

philpax:
Bah, I'm more confused now. I tried building an isolated test case to submit to R-A with a similar setup (features, wasm32-wasi) and it seems to work there. I'm guessing it's due to another factor I haven't isolated.

I'll just do the `.gitignore` thing for now, and create an issue to look into it at some later stage.

PR #1180, opened 2023-11-22
nix flake devenv

yVieta:
I am a bit new on creating flakes for developement environments. To help myself I used the [fh cli](https://github.com/DeterminateSystems/fh). I have slightly edit the missing pkgs where fh could not find.


ten3roberts:
@pierd, this may look lime something for you (assuming Docker containers skills translate to nix containers)😅

yVieta:
> @pierd, this may look like something for you (assuming Docker containers skills translate to nix containers)😅

Most docker containers are not reproducible according to [reproducible](https://reproducible-builds.org/) . Most tools  for reproducible docker containers are slightly deprecated. An example https://github.com/reproducible-containers/repro-get . Also you can also cross build via Nix --> Nix Flakes. I can add support for more architectures when needed which are all reproducible. Bit by Bit.

pierd:
Containers are about reproducible deployments so deploying the same container image gives you the same results. The actual building of an image has no guarantees.

As much as I would like to have I unfortunately have no nix experience 😅 @ten3roberts do you know if anyone on the team uses nix?

I think I'm also missing some context. @yVieta could you explain a bit why do you think it's beneficial to add this? Do I correctly assume that it is to easily build Ambient (with all dependencies like rust and alsa)?

ten3roberts:
> Containers are about reproducible deployments so deploying the same container image gives you the same results. The actual building of an image has no guarantees.
> 
> As much as I would like to have I unfortunately have no nix experience 😅 @ten3roberts do you know if anyone on the team uses nix?
> 
> I think I'm also missing some context. @yVieta could you explain a bit why do you think it's beneficial to add this? Do I correctly assume that it is to easily build Ambient (with all dependencies like rust and alsa)?

It was worth a shot. Nix would be such a good fit for server environments given their reproducible installation, compared to the "freeze the disk and ship it".

In that case, I would then be the one with the most nix experience. Have used it for a little while

yVieta:
@pierd It is much easier for contributors when they use nix development shell because all the tools for building ambient are there. They just need to use `nix develop` in the root file of ambient and have everything they need. Building and running Ambient can be done as usual via cargo. The goal of Nix development shells are to make a sustainable, reproducible and easy to modify environment to build software.

yVieta:
> > Containers are about reproducible deployments so deploying the same container image gives you the same results. The actual building of an image has no guarantees.
> > As much as I would like to have I unfortunately have no nix experience 😅 @ten3roberts do you know if anyone on the team uses nix?
> > I think I'm also missing some context. @yVieta could you explain a bit why do you think it's beneficial to add this? Do I correctly assume that it is to easily build Ambient (with all dependencies like rust and alsa)?
> 
> It was worth a shot. Nix would be such a good fit for server environments given their reproducible installation, compared to the "freeze the disk and ship it".
> 
> In that case, I would then be the one with the most nix experience. Have used it for a little while

You can also try Nix out (the package manger and the programming language) installing it next to your system. 
The configuration of [Determinate Systems](https://determinate.systems/posts/determinate-nix-installer) brings you a smoother entry to Nix and allows you to remove it anytime you want. The [official configs](https://nixos.org/manual/nixos/stable/#sec-installing-from-other-distro) are less out of the box and harder to remove from the system.

ten3roberts:
Awesome, I have used nix a little bit before on a personal laptop.

We are considering setting up a CI step for this to ensure that the flake does not go stale as we add new things.

It is most definitely a cool concept to have such a declarative environment to setup all the necessary packages in the right places, rather than a `setup.sh` that works half of the time 😅

ten3roberts:
Hi @yVieta,

Thank you for your contribution.

We do unfortunately not have enough experience with Nix, and we currently do not have sufficient need or application for a nix environment to justify maintenance cost and testing.

After discussing this with the team, we have unfortunately decided to not include this feature as a part of our codebase.

PR #1184, opened 2023-11-23
Added URL to component and field type mismatch error messages

### LINKED ISSUE #1183: Request: More didactic error message when invalid type in ambient.toml
Each time I have an invalid component type, I feel like there's a missed opportunity to teach me what the valid component types _are._

```toml
boid_neighbour_count = { type = "Usize" }
boid_neighbour_array = { type = "Array" }
```

```Error: Failed to resolve dependencies for pre-build

Caused by:
    Failed to resolve type `Usize` for component `boid_neighbour_count```

This error message could be followed with a message like the following:

```Valid types are `U8, U32, I8, I32, F32, Bool, EntityId, . . .````

droqen:
My attempt to resolve #1183 
I haven't successfully built Ambient from this, I tried and it took a super long time... but I do expect it to work. The change is very small.

### DIFF
--- a/shared_crates/package_semantic/src/component.rs
+++ b/shared_crates/package_semantic/src/component.rs
@@ -53,8 +53,9 @@ impl Resolve for Component {
                 .get_contextual_type_id(parent_id, ty)
                 .with_context(|| {
                     format!(
-                        "Failed to resolve type `{ty:?}` for component `{}`",
-                        self.data.id
+                        "Failed to resolve type `{ty:?}` for component `{}`\nSee {} for valid types",
+                        self.data.id,
+                        "https://ambientrun.github.io/Ambient/reference/package.html#valuetype"
                     )
                 })?,
             ResolvableItemId::Resolved(id) => *id,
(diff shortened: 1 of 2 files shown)


droqen:
Rather than give a list of all types (very long, dynamic), I put a link in the error message to the relevant section in the book (shorter, static, but might change). Is that something we're avoiding doing for some reason?

philpax:
Looks good to me, but we might want to use https://ambient.run/docs/reference/package instead. However... that might be out of date (it's not updated automatically with the repo - but we probably want versioning instead), and more annoyingly, it doesn't have anchors (https://github.com/AmbientRun/Web/issues/47).

@FredrikNoren Have we decided where the docs should live? If it's `ambient.run` then I'll see if I can fix up the anchors issue and we'll deal with versioning later I suppose

philpax:
Ok, thought about it some more, we should use `ambient.run`. I've fixed #47, so you should be able to replace the URL with https://ambient.run/docs/reference/package#valuetype - once you've done that, feel free to merge it!

droqen:
Oops, I missed your edit. I made an issue for it, haha. #1191

PR #1192, opened 2023-11-24
Fix package deployment failure

philpax:
Since the ~11th, package nightlies have not been deployed, and nobody was aware of them. This is a combination of two issues:

1) Failures in Campfire-spawned Ambient commands did not cause Campfire to exit with a non-zero status code, resulting in a silent failure
2) Our download script for the Unity assets was constructing URLs with double slashes (i.e. `BASE//path` instead of `BASE/path`), resulting in invalid assets being downloaded. My guess is that DO stopped accepting double-slashed URLs?

This PR fixes both of these issues, and should hopefully allow for package deployment to go out as per normal.


pierd (approved):
LGTM

Issue #1121, opened 2023-10-26
Measure end to end latency


pierd:

pierd:
https://github.com/AmbientRun/Ambient/pull/1022 adds `debug-local-datagram-latency` feature for debugging the latency of sent unreliable messages, it relies on clocks being in sync so it works the best on localhost or via proxy (with both client and server running on the same machine).

pierd:
Reopening as I'm going to explore adding true end to end latency (from user input to rendered frame).

pierd:
Latency talk to watch: https://www.gdcvault.com/play/1025968/Controller-to-Display-Latency-in

pierd:
Created a quick version in a branch: https://github.com/AmbientRun/Ambient/tree/kuba/timings

It currently spams the logs with all the measurements:
```
18:07:38  WARN ambient_timings: TIMING EVENT: Input 7.981000166s
18:07:38  WARN ambient_timings: TIMING EVENT: RenderingFinished 7.986576833s
18:07:38 ERROR ambient_timings: TIMINGS: Input <- 4.855708ms -> ScriptingStarted <- 10.20825ms -> ScriptingFinished <- 6.875µs -> ClientSystemsStarted <- 118.125µs -> ClientSystemsFinished <- 79.333µs -> DrawingWorld <- 379.709µs -> DrawingUI <- 144.083µs -> SubmittingGPUCommands <- 5.973875ms -> RenderingFinished
18:07:38 ERROR ambient_timings: INPUT TO RENDERED: 21.765958ms
18:07:38  WARN ambient_timings: TIMING EVENT: ScriptingStarted 7.986774625s
18:07:38  WARN ambient_timings: TIMING EVENT: ScriptingFinished 7.997007375s
18:07:38  WARN ambient_timings: TIMING EVENT: ClientSystemsStarted 7.997012041s
18:07:38  WARN ambient_timings: TIMING EVENT: ClientSystemsFinished 7.997129458s
18:07:38  WARN ambient_timings: TIMING EVENT: DrawingWorld 7.997216875s
18:07:38  WARN ambient_timings: TIMING EVENT: DrawingUI 7.997581333s
18:07:38  WARN ambient_timings: TIMING EVENT: SubmittingGPUCommands 7.997727625s
18:07:38  WARN ambient_timings: TIMING EVENT: RenderingFinished 8.002686125s
[...]

PR #1211, opened 2023-12-06
Add frame timings

pierd:
Addresses #1121 

Makes input to rendered available in the debugger.


crates/debugger/src/lib.rs
```diff
@@ -143,12 +156,23 @@ pub fn AppStatsView(hooks: &mut Hooks) -> Element {
             external_time.apply(sample.external_time);
         }
 
-        set_measurements((frame_time, external_time))
+        set_measurements((frame_time, external_time));
```
ten3roberts, line 159: So if I've gotten this correctly, it will use my previous `Measurement` to provide a moving average, min, and max from input processing to the end of the rendering system.

I don't seem to find a use of the `Display` impl for FrameTimings to report each frame, where is that, and is it just the latest or also a moving average?

ten3roberts (changes requested):
Left some comments and questions 😊

crates/timings/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+const MAX_SAMPLES: usize = 128;
+static ENABLED: AtomicBool = AtomicBool::new(false);
+
+pub fn set_enabled(enabled: bool) {
+    ENABLED.store(enabled, std::sync::atomic::Ordering::Relaxed);
+}
+
+/// Frame events being timed (in order!)
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, FromPrimitive, ToPrimitive)]
+pub enum TimingEventType {
```
ten3roberts, line 18: Would it be possible to make this less hardcoded and instead use a `&'static str` instead?

Right now, this creates a conceptual dependency between all of ambients system and our internals, and this sub-crate, which increases coupling even more.

Let me know if I am being too hard on the review, have been in the open-source reviewing sphere lately where they are *very* strict 😅

philpax:
Shall we merge this?

pierd:
> Shall we merge this?

I'm going to update the structure a bit to make packages less coupled but that'll be after the Xmas break.

As for the coupling itself.. there'll still be coupling around the reporting api (whatever the report method would it be other parts have to know how (currently `report_event` on `ambient_timings::Reporter`) and what to report (currently `TimingEvent`/`TimingEventType`)). Using `&'static str` just makes the coupling implicit instead of explicit.

crates/network/src/client_game_state.rs
```diff
@@ -147,10 +151,20 @@ impl ClientGameState {
             None,
         );
 
+        timings_reporter.report_event(TimingEventType::SubmittingGPUCommands);
         gpu.queue.submit(Some(encoder.finish()));
         for action in post_submit {
             action();
         }
+        let callback = move || timings_reporter.report_event(TimingEventType::RenderingFinished);
+        #[cfg(target_os = "unknown")]
```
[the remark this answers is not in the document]
pierd: Exactly

crates/timings/src/lib.rs
ten3roberts: That sounds good, I suppose this will be a case of seeing which part goes the easier, like when playing Jenga.

I'd personally go with the core approach, as adding another crate for it may not be worthwhile over the increased link time and risks of cyclic dependencies.

ten3roberts (approved):
Looks great,

I suppose we could say it is a good timing to merge this 😛

